PR #9, opened 2023-10-17
Pull requests now give CI feedback for errors in doxygen documentation

StudentKyle:
## This pull request:
- Changes WARN_AS_ERROR from NO to FAIL_ON_WARNINGS. This will make the doxygen build test fail if doxygen encounters an error or warning (such as undocumented code) while still continuing to completion
- Adds a github actions bot to comment the doxygen warnings onto the pull request. This allows us to easily see feedback when a test has undocumented members so that they can be fixed before merging to main.
- Disables latex in doxygen build, since it has a slow install. We can re-enable this later on the deploy part if we want to utilize latex features in our docs.

Keep in mind, warnings will **accumulate** if they are allowed to merge regardless, making all future pull requests also display those warnings. This behavior should be helpful to encourage us to complete our documentation fully.  

Ready for review and merging! 
## **EXCEPT:**

Do we want to get warnings for undocumented private members, such as `pin` in `RevEncoder`?
I see three options: 
**1. We require private members to be documented as well, giving warnings on undocumented code.** (current choice)
2. Suppress private member warnings on test runs, but still generate autodocs for them on builds (by utilizing two Doxyfiles).
3. (worst option, I think) We disable all autodocs for private members, meaning we won't have any doxygen documentation on our private members at all (but at least we don't get spurious warnings)

### After discussion, we now require documentation for all objects and members, public **or** private.


StudentKyle:
Cool, it failed (from undocumented members from initial commits on main)

StudentKyle:
Cool, it failed (from undocumented members from initial commits on main). Though there are also some versioning warnings.

StudentKyle:
Reopening to run actions

StudentKyle:
It works! Ready for review and merging if it looks good.

StudentKyle:
We now require public and private members to be documented after discussions.

nitroxplunge (approved):

PR #10, opened 2023-10-21
Only accept documentation from src/ and docs/

StudentKyle:
Now we will only attempt to document _our_ code and not external resources.


StudentKyle:
It worked. Those errors shown above are from our previously existing code in src/, not the external teensy code

nitroxplunge (approved):

PR #11, opened 2023-10-25
Add githook-provided branch protection

StudentKyle:
Adds a pre-commit githook to stop commits from happening on main locally. 

The githook(s) are located at .githooks/ . This directory needs to be set individually on each repository clone so the hook is tracked here on the repository, and then run by each person's local git.

This is set by 
```bash
git config --local core.hooksPath .githooks
```

I have put this in the main README.md, and in addition an init task in `Makefile-githooks`. In the future we should integrate it with a single Makefile setup/init task, but I do not know how our build script will work yet.

### DIFF
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -1,0 +1,19 @@
+#!/bin/bash
+
+# pre-commit githook, runs before a git commit is accepted.
+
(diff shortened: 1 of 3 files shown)


Rishi-H (approved):
Good

PR #14, opened 2023-10-28
Feature rev encoder

AidanDonnelly1:


Rishi-H (approved):

Rishi-H:
do better next time aidan

src/sensors/rev_encoder.hpp
```diff
@@ -1,16 +1,23 @@
...
+ * @file rev_encoder.hpp
+ * @brief Header file for the Rev Encoder 
+ * @date 2023-10-14
+ */
+#include <cmath>
 #include <FreqMeasureMulti.h>
 
-#ifndef REV_ENCODER_H
-#define REV_ENCODER_H
+#ifndef RevEncoder_H
```
StudentKyle, line 9: This should be all caps I believe

src/sensors/rev_encoder.hpp
```diff
@@ -1,16 +1,23 @@
...
-		uint8_t pin;
+		uint8_t inPin;
 		FreqMeasureMulti freq;
 	public:
-		RevEncoder(uint8_t pin);
-		int getRawOutput();
-		float getAngle();
+		RevEncoder(uint8_t encPin);
+        void read();
+		float ticks, radians;
```
StudentKyle, line 20: You have undocumented members and objects, please document these

src/sensors/rev_encoder.cpp
```diff
@@ -1,18 +1,46 @@
-#include "rev_encoder.hpp"
+/**
+ * @file rev_encoder.cpp
+ * @brief The implementation of the Rev Encoder 
+ * @date 2023-10-14
+ * 
+ * @copyright Copyright (c) 2023
+ * 
+ */
```
StudentKyle, line 8: We use triple /// for our doxygen documentation. (see docs/README.md standards)

StudentKyle (changes requested):
You have documentation issues to resolve

AidanDonnelly1:
doxy gen geeking

src/sensors/rev_encoder.hpp
```diff
@@ -1,16 +1,58 @@
...
+
+#include <cmath>
 #include <FreqMeasureMulti.h>
 
 #ifndef REV_ENCODER_H
 #define REV_ENCODER_H
 
-class RevEncoder {
+/**
+ * @brief the class for the Rev Encoder
```
StudentKyle, line 14: Can you describe what a "rev encoder" is instead? What is it, why is it useful? 
Also, please use triple /// instead for all doxygen comments

src/sensors/rev_encoder.hpp
```diff
@@ -1,16 +1,58 @@
...
+		*/
+		uint8_t inPin;
+		/**
+		* @brief Used to read rise time of the encoder
+		*/
 		FreqMeasureMulti freq;
+		/**
+		* @brief measure of current angle in ticks (1024 ticks = 1 rotation). When set to 1 individual readings are returned
+		*/
+		float ticks;
```
StudentKyle, line 30: Make sure your software doesn't reduce its data resolution by restricting it to 1024 integers like we discussed earlier. Just check what resolution it supports, and as long as you don't do any int casts you should be fine

src/sensors/rev_encoder.hpp
```diff
@@ -1,16 +1,58 @@
...
-		int getRawOutput();
-		float getAngle();
+		/**
+		* @brief Construct a new rev_encoder object
+		* @param encPin the pin number that the encoders signal pin is plugged into
+		* @param baudrate the desired baudrate for the encoder
+		*/
+		RevEncoder(uint8_t encPin, int baudrate);
+		/**
+		* @brief updates the current angle in ticks (1024 = 1 Rotation) and radians.
```
StudentKyle, line 43: What is the difference between ticks and radians? If there is no difference, do we need both?

Also, remember to switch to /// (we do this so we can more independently change lines of doxygen comments, and it makes it explicit what is doxygen vs what is not.

StudentKyle (changes requested):
Have a few comments on your documentation. Most of it looks great though (but see comments)!

src/sensors/rev_encoder.hpp
AidanDonnelly1: added a link to the exact item

src/sensors/rev_encoder.hpp
AidanDonnelly1: yes

src/sensors/rev_encoder.hpp
AidanDonnelly1: it's up to rishi if he doesn't want both I can remove one.

src/sensors/rev_encoder.hpp
StudentKyle: Also, it's worth including a note about the exact parts/model numbers we use in the robots

src/sensors/rev_encoder.hpp
```diff
@@ -1,57 +1,33 @@
...
+		/// @brief measure of current angle in radians 
 		float radians;
 	public:
-		/**
-		* @brief Construct a new rev_encoder object
-		* @param encPin the pin number that the encoders signal pin is plugged into
-		* @param baudrate the desired baudrate for the encoder
-		*/
+		/// @brief Construct a new rev_encoder object
+		/// @param encPin the pin number that the encoders signal pin is plugged into
```
StudentKyle, line 21: encPin should be snake_case (also I suggest not shortening variable names if you can help it, I.e. encoder_pin instead)

src/sensors/rev_encoder.hpp
```diff
@@ -1,16 +1,34 @@
...
-		float getAngle();
+		/// @brief Construct a new rev_encoder object
+		/// @param encPin the pin number that the encoders signal pin is plugged into
+		/// @param baudrate the desired baudrate for the encoder
+		RevEncoder(uint8_t encPin, int baudrate);
+		/// @brief updates the current angle in ticks (1024 = 1 Rotation) and radians.
+		void read();
+		/// @brief get the last angle of the encoder ticks
+		/// @return the last angle of the encoder in ticks (1024 = 1 Rotation).
+		float getAngleTicks();
```
StudentKyle, line 28: snake_case
```suggestion
		float get_angle_ticks();
```

src/sensors/rev_encoder.hpp
```diff
@@ -1,16 +1,34 @@
...
+		/// @brief measure of current angle in radians 
+		float radians;
 	public:
-		RevEncoder(uint8_t pin);
-		int getRawOutput();
-		float getAngle();
+		/// @brief Construct a new rev_encoder object
+		/// @param encPin the pin number that the encoders signal pin is plugged into
+		/// @param baudrate the desired baudrate for the encoder
+		RevEncoder(uint8_t encPin, int baudrate);
```
StudentKyle, line 23:
```suggestion
		RevEncoder(uint8_t encoder_pin, int baudrate);
```

src/sensors/rev_encoder.hpp
```diff
@@ -1,16 +1,34 @@
...
+		/// @brief updates the current angle in ticks (1024 = 1 Rotation) and radians.
+		void read();
+		/// @brief get the last angle of the encoder ticks
+		/// @return the last angle of the encoder in ticks (1024 = 1 Rotation).
+		float getAngleTicks();
+		/// @brief get the last angle of the encoder in radians
+		/// @return the last angle of the encoder in radians
+		float getAngleRadians();
```
StudentKyle, line 31: snake_case
```suggestion
		float get_angle_radians();
```

src/sensors/rev_encoder.cpp
```diff
@@ -1,18 +1,29 @@
...
+      this->radians = (this->ticks / 1024) * M_PI * 2;
+    }
 }
 
-float RevEnc::getAngle() {
-	int rawAngle = this->getAngleRaw();
-	return rawAngle == -1 ? -1 : map(rawAngle, 1, 1024, -PI*1000.0, PI*1000.0) / 1000.0;
-}
\ No newline at end of file
+float RevEncoder::getAngleTicks()
```
StudentKyle, line 21:
```suggestion
float RevEncoder::get_angle_ticks() 
```

src/sensors/rev_encoder.hpp
```diff
@@ -1,16 +1,34 @@
...
 #define REV_ENCODER_H
 
-class RevEncoder {
+/// @brief the class for the Rev Through Bore Encoder(www.revrobotics.com/rev-11-1271/)
+class RevEncoder 
+{
 	private:
-		uint8_t pin;
+		/// @brief the pin number that the encoders signal pin is plugged into
+		uint8_t inPin;
```
StudentKyle, line 12:
```suggestion
		uint8_t in_pin;
```

src/sensors/rev_encoder.cpp
```diff
@@ -1,18 +1,29 @@
-#include "rev_encoder.hpp"
+#include <rev_encoder.hpp>
 
-RevEnc::RevEnc(uint8_t pin) {
-    this->pin = pin;
-	freq.begin(this->inPin, FREQMEASUREMULTI_MARK_ONLY);
+RevEncoder::RevEncoder(uint8_t encPin, int baudrate)
```
StudentKyle, line 3:
```suggestion
RevEncoder::RevEncoder(uint8_t encoder_pin, int baudrate) 
```

src/sensors/rev_encoder.cpp
```diff
@@ -1,18 +1,29 @@
-#include "rev_encoder.hpp"
+#include <rev_encoder.hpp>
 
-RevEnc::RevEnc(uint8_t pin) {
-    this->pin = pin;
-	freq.begin(this->inPin, FREQMEASUREMULTI_MARK_ONLY);
+RevEncoder::RevEncoder(uint8_t encPin, int baudrate) 
+{
+  this->inPin = encPin;
```
StudentKyle, line 5:
```suggestion
  this->in_pin = encoder_pin;
```

src/sensors/rev_encoder.cpp
```diff
@@ -1,18 +1,29 @@
...
-	while (this->freq.available() > 1) this->freq.read(); // Burn through buffer of values in freq
-	int angle = round(freq.countToNanoseconds(freq.read()) / 1000.0);
-	if (angle >= 1 && angle <= 1025) return angle;
-	return -1;
+void RevEncoder::read() 
+{
+  if (freq.available()) 
+    {
+      float frequency = this->freq.countToNanoseconds(this->freq.read()) / 1000;
+      this->ticks = (int) frequency % 1024;
```
StudentKyle, line 16: Here you've done the int cast I was talking about. Can you convert this into a normalized float instead so that this doesn't reduce our resolution?

StudentKyle (changes requested):
Functions and variables should be snake_case. There's the int cast I was talking about present in your code.

I am still wondering about the data format that we get from your get_angle_ticks() and get_angle_radians(). What value is "0" rotation, what does that mean physically? What about the max rotation (what value, is that a full revolution)?

And what is the difference between a tick and a radian? If there's no difference (i.e. they measure the same thing), do we need both to be public?

src/sensors/rev_encoder.hpp
StudentKyle: You have an int cast, see below review

src/sensors/rev_encoder.hpp
StudentKyle: Do they measure the same thing? If they do, it would be better to remove one (or if you use the less-intuitive one internally, keep it private, while keeping the intuitive output getter public)

src/sensors/rev_encoder.hpp
StudentKyle, line 31: What is the angle (tick or radian) measured from?

- What is 0 rotation?
- What value represents one full revolution?
- What happens after a full revolution? Does it keep counting up, jump back to 0, etc?
- Can there be negative numbers? Does positive values represent clockwise or counterclockwise rotation?

(if each of these questions can be answered in the documentation that'd make using your values much easier.

src/sensors/rev_encoder.cpp
StudentKyle: Just checked the docs you linked, the sensor has a resolution of:

"Quadrature Resolution: 2048 Cycles per Revolution (8192 Counts per Revolution)"

if you switch your 1024 to 8192 we should maintain all the resolution (make sure your frequency aligns with the new 1 revolution as well)

src/sensors/rev_encoder.hpp
StudentKyle: After the discussion, you could answer these questions by describing something like:
on your get_ticks, returns in the range of [0,1023], where 1024 ticks = 2pi radians, (wrapped between revolutions)
get_radians returns in range [0,2pi] radians (wrapped between revolutions)

src/sensors/rev_encoder.cpp
StudentKyle: Or maybe I don't understand the sensor: it's 8192 incremental outputs, but 1024 absolute frequency outputs? I'll probably leave it to you then. Either way you should be good to have ticks as an int.

StudentKyle (approved):
Looks good to me

PR #15, opened 2023-11-01
[WIP] Adds IMU Sensor Support

StudentKyle:
Tested and working! LSM on I2C, and ICM on I2C and SPI 

Adds support for the IMU Sensors ICM20649 and LSM3DSOX+LIS3MDL.

The ICM20649 has support for acceleration and gyroscope, in SPI or I2C mode (set on running init() ), for 3+3=6 degrees of freedom

The LSM3DSOX+LIS3MDL supports acceleration, gyroscope, and magnetometer readings (9 degrees of freedom) but software only supports the acceleration and gyroscope. This LSM model only supports I2C , NOT SPI.

Simply make the object, call init() to initialize once, then read() every time you want to read, then you can access the data (updated after read() ) from the getters.


src/sensors/imu/LSM6DSOX.cpp
```diff
@@ -0,0 +1,33 @@
+#include "LSM6DSOX.hpp"
+
+LSM6DSOX::LSM6DSOX() {}
+
+// initialize LSM
+void LSM6DSOX::init() {
+    // start I2C communication 
+    sensor.begin_I2C();
+    // TODO switch to SPI
```
Pandabear1125, line 9: Bruh

Pandabear1125 (changes requested):

src/sensors/imu/LSM6DSOX.cpp
StudentKyle: fixed

StudentKyle:
I performed an in-person review with @nitroxplunge, said it looks good after changing the examples directory name.

PR #16, opened 2023-11-01
Downgrade doxygen to version provided in linux packages

StudentKyle:
From version 1.9.5 to version 1.9.1, removing bounding box errors


Pandabear1125 (approved):
Looks good (I'm blind)

PR #19, opened 2023-11-15
Add Safety Code

LachlanMurphy:
\_(*0*)_/

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -1,12 +1,20 @@
 #include "utils/timing.h"
+#include "sensors/dr16.hpp"
+#include "comms/rm_CAN.hpp"
 
 uint32_t cycle_time_us = 1000;
 uint32_t cycle_time_ms = cycle_time_us / 1000;
 float cycle_time_s = cycle_time_us * 1E-6;
 
+DR16 dr16;
+rm_CAN can;
+
 // Runs once
 void setup() {
 	Serial.begin(1000000); // the serial monitor is actually always active (for debug use Serial.println & tycmd)
+
+    dr16.init(); // set up dr16
+    can.init(); // set up can
 
 	if (Serial) {
 		Serial.println("TEENSY SERIAL START\n\n");
@@ -42,6 +50,8 @@ void setup() {
 
 // Master loop
 int main() { // Basically a schudeling algorithm
+    dr16.read(); // read data from controller
+
     Timer timer;
 
 	unsigned long prev_time = micros();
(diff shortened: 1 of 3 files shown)


StudentKyle:
Super excited about safety code, thanks for doing this. What's the plan to safely test the safety code?

The docs errors shown in that comment include some of your code, including the keys struct (as a whole), keys members, and m_disctTime. 

Definitely document keys (the base member struct containing all the keys) and disctTime (note snake_case).

In this case I think the keys members are pretty self-explanatory (would match behavior described on keys member). You can document each key letter if you'd like (great if you do!), but I found a way to manually suppress warnings within a range if you don't feel like doing it on this specific case.

https://stackoverflow.com/questions/16993250/suppress-doxygen-warning-for-undocumented-member-function-but-leave-synopsis-in#29490878

(I suggest using the alias \nowarn and \endnowarn)

Let's only do this if docs are REALLY clear to not just the author but other people as well. I think a bool for whether an individual key is pressed is as simple as it gets as long as the struct containing it is documented that describes the containing behavior. 

(this might not be the case for something even slightly more complex, like a struct containing each joystick analog output. Each one's output range should be be described in that hypothetical case. So let's make sure to ask what others think about docs if we ever want to suppress warnings)

StudentKyle:
Though maybe we'd want to document them anyway so we can see its description, even if it's just "whether or not w is pressed" during auto complete. That could be nice, and it would avoid needing to make a distinction for what is self-explanatory or not - which could save time and questions in the long run.

LachlanMurphy:
I'll document the code Tuesday during meeting. As for testing it might be a little bit tricky considering we haven't compiled without an .ino file yet. I'll work on making that happen after documentation. As for when we can test I'm thinking we just isolate one motor in a controlled environment (on the floor on blocks) and try to break the code. I'll let you know about specifics when I get closer to the point.

StudentKyle:
I believe @saaaji and @nitroxplunge have plans for the build script. For now just using the .ino like you said is fine that just tests your code (you can make one yourself that creates the class and runs the code you want to test), but you can talk with them about the full integration pipeline if you want.

LachlanMurphy:
@StudentKyle Everything has been documented. Jackson and I (separately) tested the safety code and we both found it to not break. Jackson found a small thing that didn't matter too much with the order for checks for the remote still sending data and the safety switch. We can also do some tests with your supervision if you would like.

StudentKyle (approved):
Docs look good

PR #22, opened 2023-12-18
Merge makefile 

Pandabear1125:


Pandabear1125:
Makefile is functional (and easy to use), ready to be used :)

libraries/Adafruit_ICM20X/.DS_Store
StudentKyle, line 1: .DS_Store is a MacOS file structure thing that isn't needed. We should probably remove it, likely by adding it to the gitignore (there are .DS_Store files elsewhere as well)

StudentKyle (changes requested):
Add .DS_Store to gitignore

Rishi-H (approved):
Looks good to me. Thanks for spending the time to do this.

Pandabear1125:
Keeping branch open in case makefile requires revision.

PR #36, opened 2024-03-10
Feature d200 lidar

saaaji:
basic d200 lidar driver is up and running. works in arduino IDE but haven't been able to see if it builds with the rest of the codebase.


src/sensors/d200.hpp
```diff
@@ -0,0 +1,131 @@
...
+
+    /// @brief stop the LiDAR
+    void stop();
+
+    /// @brief read latest packet from LiDAR module
+    void read();
+
+    /// @brief get the latest packet
+    /// @return the latest packet
+    LidarDataPacket get_latest_packet();
```
StudentKyle, line 125: What happens if there is not new data to be read? Does it give duplicate packets (ideally labeled with the old teensy timestamp)? Or something else? Please document this behavior as well as make sure it's correct.

src/sensors/d200.hpp
```diff
@@ -0,0 +1,131 @@
...
+  struct {
+    /// @brief distance for point (mm)
+    uint16_t distance;
+    /// @brief signal strength for point (normalized)
+    uint8_t intensity;
+  } points[POINTS_PER_PACKET];
+  /// @brief end angle for measurements (0.01 deg)
+  uint16_t end_angle;
+  /// @brief timestamp of start of measurements (ms, wraps after 30000)
+  uint16_t timestamp;
```
StudentKyle, line 82: Can you instead just throw out this data? This should be assigned to the teensy time instead (through micros() or millis(), but always converted into seconds for the final return)

src/sensors/d200.hpp
```diff
@@ -0,0 +1,131 @@
...
+/// @brief default scanning speed (deg/s)
+const uint16_t DEFAULT_SPEED = 6 * 360;
+/// @brief minimum specified scanning speed (deg/s)
+const uint16_t MIN_SPEED = 2 * 360 + 1;
+/// @brief maximum specified scanning speed (deg/s)
+const uint16_t MAX_SPEED = 8 * 360 - 1;
+
+/// @brief struct for individual LiDAR packet
+typedef struct {
+  /// @brief current speed of LiDAR (deg/s)
```
StudentKyle, line 68: Convert all units of degrees into radians, all units of time into seconds

src/sensors/d200.hpp
```diff
@@ -0,0 +1,131 @@
...
+  uint16_t lidar_speed;
+  /// @brief start angle for measurements (0.01 deg)
```
StudentKyle, line 70: convert to radians

src/sensors/d200.hpp
```diff
@@ -0,0 +1,131 @@
...
+  uint16_t start_angle;
+  /// @brief distance and intensity measurements
+  struct {
+    /// @brief distance for point (mm)
+    uint16_t distance;
```
StudentKyle, line 75: convert distances into meters

src/sensors/d200.hpp
```diff
@@ -0,0 +1,131 @@
...
+    /// @brief signal strength for point (normalized)
+    uint8_t intensity;
+  } points[POINTS_PER_PACKET];
+  /// @brief end angle for measurements (0.01 deg)
```
StudentKyle, line 79: convert to radians

src/sensors/d200.hpp
```diff
@@ -0,0 +1,131 @@
...
+  /// @brief current speed of LiDAR (deg/s)
+  uint16_t lidar_speed;
+  /// @brief start angle for measurements (0.01 deg)
+  uint16_t start_angle;
+  /// @brief distance and intensity measurements
+  struct {
+    /// @brief distance for point (mm)
+    uint16_t distance;
+    /// @brief signal strength for point (normalized)
+    uint8_t intensity;
```
StudentKyle, line 77: Normalized into what? What is the expected range? Does a low value represent a low or high intensity?

src/sensors/d200.hpp
StudentKyle, line 125: Make sure the end result is converted into radians, meters, and seconds, since that is what the rest of the project expects from every sensor.

You could have LidarDataPacket be the raw data as you currently have parsed, and then return a different struct with the filtered data into converted units. Could also be a good post-process step to throw out sensor time and give the teensy time instead (in seconds). 

**Addendum:** Make sure the timestamp is set when the complete packet is read, NOT when gotten in the getter here.

src/sensors/d200.hpp
```diff
@@ -0,0 +1,131 @@
...
+    /// @param buf pointer to buffer
+    /// @param len length of buffer
+    /// @return CRC8 checksum for buffer
+    uint8_t calc_checksum(uint8_t *buf, int len);
+
+  public:
+    /// @brief constructor and initialization
+    D200LD14P();
+
+    /// @brief set the speed of the LiDAR
```
StudentKyle, line 110:
```suggestion
    /// @brief set the rotation speed of the LiDAR
```

src/sensors/d200.cpp
```diff
@@ -0,0 +1,103 @@
...
+        p->start_angle = (b[5] << 8) | b[4];
+
+        for (int i = 0; i < POINTS_PER_PACKET /* 12 */; i++) {
+          int base = 6 + i * 3;
+          p->points[i].distance = (b[base + 1] << 8) | b[base];
+          p->points[i].distance = b[base + 2];
+        }
+
+        p->end_angle = (b[43] << 8) | b[42];
+        p->timestamp = (b[45] << 8) | b[44];
```
StudentKyle, line 62: Throw out timestamp from lidar, instead use teensy time. Can be done here or in a separate post-process step. I'd prefer a separate post-process step that happens as soon as the read() function reads a whole packet, that can also do things like unit conversions into a result struct.

src/sensors/d200.cpp
```diff
@@ -0,0 +1,103 @@
...
+  uint8_t msb = (speed >> 8) & 0xff;
+
+  uint8_t cmd[CMD_LEN] = { 0x54, 0xa2, 0x04, lsb, msb, 0, 0, 0 };
+  cmd[CMD_LEN - 1] = calc_checksum(cmd, CMD_LEN - 1);
+
+  LIDAR_SERIAL.write(cmd, CMD_LEN);
+}
+
+void D200LD14P::start() {
+  LIDAR_SERIAL.write(START_CMD, CMD_LEN);
```
StudentKyle, line 32: Does this start/stop it from spinning?

src/sensors/d200.cpp
```diff
@@ -0,0 +1,103 @@
...
+}
+
+void D200LD14P::stop() {
```
StudentKyle, line 35: Does this start/stop it from spinning?

src/sensors/d200.cpp
```diff
@@ -0,0 +1,103 @@
...
+  LIDAR_SERIAL.write(STOP_CMD, CMD_LEN);
+}
+
+void D200LD14P::read() {
+  while (LIDAR_SERIAL.available()) {
```
StudentKyle, line 40: I really like how you handled this to support any length of serial buffer. However, two edge cases to fix/verify:

**Case 1:**
1. You have a partially filled buffer, with say, 20 bytes in it, from a previous read() call, and are awaiting the next 27 bytes.
2. Time passes such that LIDAR_SERIAL gets full and overwrites old data multiple times. 
    Now, the partially filled packet would be invalid to continue filling, but code doesn't know that yet.
3. Read() is called some time later. Maybe a new start byte is encountered, maybe not, but regardless your code currently will not reset the buffer index = 0. Instead, it will read the next 27 garbage bytes as valid data. The checksum might catch it, but this isn't guarenteed since measurement data section could be anything (unless I am misunderstanding the checksum standard).
This edge case looks broken to me. 

I think this could be fixed by checking if LIDAR_SERIAL.available() == MAX_SERIAL_BUFFER. If true, it is possible something was overwritten. So, throw out the in-progress packet (reset idx = 0) and begin checking for a new start character instead.
This solution comes at a cost of lost packets, but I think correctness is more important here since this shouldn't happen too often unless the main loop code is too slow.


**Case 2:**
LIDAR_SERIAL has more than PACKET_SIZE bytes, so two possibilities exist:
there is one complete packet and one incomplete packet, or two incomplete packets. 

I ran through possibilities in my head and it looks like your solution functions, but you should make sure this performs as expected (don't repeatedly drop packets, etc).

src/sensors/d200.hpp
```diff
@@ -0,0 +1,131 @@
...
+  /// @brief most recently read packet
+  LidarDataPacket latest_packet;
+  /// @brief buffer to assemble next packet
+  uint8_t buffer[PACKET_SIZE];
+  /// @brief current write index for buffer
+  uint8_t buffer_idx;
+} LidarState;
+
+/// @brief class for LiDAR driver
+class D200LD14P {
```
StudentKyle, line 96: Consider adding a function that converts a LidarPacket into a point cloud (list of x,y(,z) points), using angle + distance transformation. Perhaps this isn't needed on the teensy side, but it could be useful on the Khadas.

StudentKyle (changes requested):
Reviewed the code. Make sure to convert to the correct units, use the teensy timestep, and handle the edge case I described in one of the comments. There's also some smaller code quality things to fix.
 
Other than that I think you came up with a very clean solution to handle the lidar reading!

nitroxplunge (changes requested):
Improvements can be made when reading data

src/sensors/d200.cpp
nitroxplunge: I disagree with this.

src/sensors/d200.cpp
```diff
@@ -0,0 +1,103 @@
...
+      if (crc8 == calc_crc8) {
+        // data values are little-endian
+        uint8_t *b = &state.buffer[0];
+        LidarDataPacket *p = &state.latest_packet;
+        
+        p->lidar_speed = (b[3] << 8) | b[2];
+        p->start_angle = (b[5] << 8) | b[4];
+
+        for (int i = 0; i < POINTS_PER_PACKET /* 12 */; i++) {
+          int base = 6 + i * 3;
```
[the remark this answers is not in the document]
nitroxplunge: Seems inherit to the data structure in which case the magic numbers are fine, but they require a comment for sure

src/sensors/d200.cpp
nitroxplunge: Nope nope nope. Use the native time of the sensor. Using teensy time will result in synchronization issues

src/sensors/d200.hpp
```diff
@@ -0,0 +1,131 @@
...
+};
+
+/// @brief serial port used by LiDAR
+#define LIDAR_SERIAL Serial1
+
+/// @brief baud rate of LiDAR module
+#define LIDAR_BAUD 230400
+
+/// @brief size of a LiDAR packet (bytes)
+const int PACKET_SIZE = 47;
```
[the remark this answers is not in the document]
nitroxplunge: This is fine

src/sensors/d200.cpp
StudentKyle: @nitroxplunge The sensor time wraps around at 30s, making it basically useless on the scale of a match. Most simple is to switch to teensy time, and we don't need to handle what happens when we go from 29.9s to 0.1s. 

Correct me if I'm wrong, but "lidar time" doesn't give us any more useful information than teensy time does. And it has the extra limitation of wrapping at 30s, so we just use the better option instead.

src/sensors/d200.cpp
StudentKyle: You disagree with the edge case existing?

StudentKyle:
Note you also have a couple undocumented members (state)

src/sensors/d200.hpp
```diff
@@ -0,0 +1,131 @@
...
+const uint8_t STOP_CMD[] = { 0x54, 0xa1, 0x04, 0, 0, 0, 0, 0x4a };
+/// @brief default scanning speed (deg/s)
+const uint16_t DEFAULT_SPEED = 6 * 360;
+/// @brief minimum specified scanning speed (deg/s)
+const uint16_t MIN_SPEED = 2 * 360 + 1;
+/// @brief maximum specified scanning speed (deg/s)
+const uint16_t MAX_SPEED = 8 * 360 - 1;
+
+/// @brief struct for individual LiDAR packet
+typedef struct {
```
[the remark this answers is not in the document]
StudentKyle: resolved

src/sensors/d200.hpp
```diff
@@ -0,0 +1,131 @@
...
+    void set_speed(uint16_t speed);
+
+    /// @brief start the LiDAR
+    void start();
+
+    /// @brief stop the LiDAR
+    void stop();
+
+    /// @brief read latest packet from LiDAR module
+    void read();
```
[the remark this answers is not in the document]
StudentKyle: latest packet saved

src/sensors/d200.hpp
StudentKyle: not necessary for lidar driver, but processing will need to happen before being able to use data.

src/sensors/d200.cpp
```diff
@@ -0,0 +1,103 @@
...
+    crc8 = CRC_TABLE[(crc8 ^ *buf++) & 0xff];
+  }
+  return crc8;
+}
+
+void D200LD14P::set_speed(uint16_t speed) {
+  uint8_t lsb = speed & 0xff;
+  uint8_t msb = (speed >> 8) & 0xff;
+
+  uint8_t cmd[CMD_LEN] = { 0x54, 0xa2, 0x04, lsb, msb, 0, 0, 0 };
```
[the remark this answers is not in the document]
StudentKyle: resolved

src/sensors/d200.cpp
StudentKyle: Still need to figure this out.

src/sensors/d200.cpp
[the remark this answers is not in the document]
StudentKyle: resolved

src/sensors/d200.cpp
```diff
@@ -0,0 +1,103 @@
...
+    // if we are reading the first two bytes of a packet, check for start and frame characters
+    // otherwise, if the buffer is not full, keep filling it
+    int next_byte = LIDAR_SERIAL.read();
+    switch (state.buffer_idx) {
+      case 0:
+        if (next_byte == START_CHAR) {
+          state.buffer[state.buffer_idx++] = next_byte;
+        }
+        break;
+      case 1:
```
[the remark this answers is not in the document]
StudentKyle: discussed in meeting

src/sensors/d200.hpp
StudentKyle: Still need to know. Making new comment on non-outdated version.

src/sensors/d200.hpp
StudentKyle: Still need to figure out if "lidar time" from the packet actually gives useful information or not, or if teensy time should be used instead.

src/sensors/d200.hpp
StudentKyle: resolved

src/sensors/d200.hpp
StudentKyle: resolved

src/sensors/d200.hpp
StudentKyle: resolved

src/sensors/d200.hpp
```diff
@@ -0,0 +1,157 @@
...
+  float lidar_speed = 0.0;
+  
+  /// @brief start angle of measurements (rad)
+  float start_angle = 0.0;
+  
+  /// @brief array of point measurements
+  struct {
+    /// @brief distance (m)
+    float distance = 0.0;
+    /// @brief intensity of measurement. units are ambiguous (not documented)
```
StudentKyle, line 89: I read the docs. It says: 

"the higher the intensity, the larger the signal strength value; the lower the intensity, the smaller the signal strength value." Can this be added?

```suggestion
    /// @brief intensity of measurement. The higher the intensity, the larger the signal strength value.
```

StudentKyle: It's likely it's just bound by the 1-byte limit unless you found otherwise via testing.

src/sensors/d200.hpp
```diff
@@ -0,0 +1,157 @@
...
+    uint8_t intensity = 0;
+  } points[D200_POINTS_PER_PACKET];
+  
+  /// @brief end angle of measurements (rad)
+  float end_angle = 0.0;
+  
+  /// @brief timestamp of measurements, wraps after 30s (s)
+  float timestamp = 0.0;
```
StudentKyle, line 97: We still need to figure out what to do with the timestamp. Since it wraps at 30s, the value provided from the lidar would not be useful over the course of a match without additional processing (increase wrap range from this provided time, or just use teensy time).

For WorldMap I need a timestamp that is reliable over the entire course of a match, and be one that I can compare with a current time (either lidar or teensy current time). However you think that can happen is fine, but if we use teensy time we won't need to send an additional timestamp representing "current time" just for lidar through comms. 

(Regardless we'll need to send some kind of lidar timestamp through so the points have some time associated with them)

src/sensors/d200.hpp
```diff
@@ -0,0 +1,157 @@
...
+    /// @return pointer to start of packet array
+    LidarDataPacket *get_packets() { return packets; }
+
+    /// @brief get the index of the most recent packet
+    /// @return index of latest packet
+    int get_latest_packet_index() { return current_packet; }
+
+    /// @brief get the latest packet
+    /// @return the latest packet
+    LidarDataPacket get_latest_packet() { return packets[current_packet]; }
```
StudentKyle, line 151: What happens if there are no packets ready? Does it return the "previous" latest packet? This behavior should be documented. 

(Currently it looks like it'll crash if the array is empty. Could the array be empty during normal execution (if packet array is cleared)? What about at the very start of execution?).

src/sensors/d200.hpp
StudentKyle: See new comment.

src/sensors/d200.cpp
StudentKyle: resolved

src/sensors/d200.cpp
StudentKyle: resolved.

src/sensors/d200.cpp
StudentKyle: Code was restructured, this analysis no longer applies

StudentKyle (changes requested):
Updated standing issues from previous review.

src/sensors/d200.hpp
```diff
@@ -0,0 +1,157 @@
...
+/// @brief default scanning speed (rad/s)
+const float D200_DEFAULT_SPEED = (float)(6 * 360) * M_PI / 180.0;
+
+/// @brief min specified scanning speed (rad/s)
+const float D200_MIN_SPEED = (float)(2 * 360 + 1) * M_PI / 180.0;
+
+/// @brief max specified scanning speed (rad/s)
+const float D200_MAX_SPEED = (float)(8 * 360 - 1) * M_PI / 180.0;
+
+struct LidarDataPacket {
```
StudentKyle, line 78: Undocumented type

StudentKyle (changes requested):
Undocumented type

src/sensors/d200.hpp
siraajsandhu: won't crash, the packets are zero initialized but I agree the behavior can be better documented.

saaaji: won't crash, the packets are zero initialized but I agree the behavior can be better documented.

src/sensors/d200.hpp
StudentKyle: Another option is to say timestamps wrapping is just fine, and the WorldMap will then handle the wrap on its level:
```python
if (new_lidar_timestamp < prev_lidar_timestamp) lidar_time =  new_lidar_timestamp + LIDAR_TIMESTAMP_WRAP
```

I don't think that kind of behavior belongs in WorldMap. It may be cleaner if all the sources of data (realsense, robot_state, lidar) just have their own, accurate timestamp handling the wrap themselves.

StudentKyle: What about a khadas-level lidar interface, that could perform the timestamp unwrapping as well as constructing the structured Point Cloud? I'd be fine with that and then it doesn't have to be a part of the lidar driver in this PR.

Rishi-H (approved):

nitroxplunge (approved):

StudentKyle (changes requested):
You have undocumented members. Please see doxygen warning output

StudentKyle:
Overall it looks good, but I can't comment on the functionality of the packet interpretation in particular. If it works and the latest packet is accessible as a LidarPacketData then good, but this should be tested.

StudentKyle:
I am still not convinced we want to be taking the lidar-packet-provided timestamp over the teensy's time, since if the lidar ever stops, we won't be able to correctly compare age on the khadas side.

Please make this decision (whichever you end up providing in the return) clear in the documentation: are we getting a lidar packet time that wraps at 30s, or is the timestamp the same one as the rest of the teensy?

StudentKyle:
After speaking with @Pandabear1125, hive can operate on the assumption the lidar won't stop. If this assumption fails, this can be a problem for later since fixing it might cost Comms bandwidth. 

I'll approve my review after the docs are fixed, and defer the testing to another reviewier. Also, Jackson should review the export_data for Comms, since 8*45 bytes is likely too much for Comms.

src/sensors/d200.hpp
```diff
@@ -0,0 +1,180 @@
...
+    /// @return the most recently read packet. if no packets have been read, this will return a zero-initialized packet
+    LidarDataPacket get_latest_packet() { return packets[current_packet]; }
+
+    /// @brief print the most recently read (complete) packet for debugging purposes
+    void print_latest_packet();
+
+    /// @brief flush the packet buffer
+    void flush_packet_buffer();
+
+    /// @brief export LiDAR data as byte array for comms
```
StudentKyle, line 175:
```suggestion
    /// @brief export LiDAR data as byte array for comms. Exports the latest D200_NUM_PACKETS_CACHED packets, for a total size of D200_NUM_PACKETS_CACHED * D200_PAYLOAD_SIZE bytes per export.
```

StudentKyle: clarify how large a byte array is given in documentation.

StudentKyle (changes requested):

StudentKyle (approved):
lgtm. We can hash out any problems via testing on the robot

PR #38, opened 2024-03-26
Feature controllers

Rishi-H:
The purpose of this branch was originally to write controllers to make the robot move. To test them we needed estimators so we made those, then we realized we needed ref and can changes to make the estimator work so we added those. Then we realized we needed a bunch of arrays (will be removed once config packet is done) to make the controllers work so we added those. We now have this monstrosity of a branch with many big changes, but in reality this working fimrware v1.0, as we finally have full state estimation and precise robot control (a little bit of drift in the imu never hurt anybody right?). Mouse and keyboard control also works, and our target states are governed before sent to the controllers. For questions about the math involved with state estimation please ask sam.


Pandabear1125 (approved):
Looks good! Great job!

PR #39, opened 2024-04-20
Fixes comms first byte problem. Fixes spelling. Adds some nice print …

Pandabear1125:
…functions

### DIFF
--- a/src/comms/usb_hid.cpp
+++ b/src/comms/usb_hid.cpp
@@ -7,7 +7,7 @@ uint16_t CommsPacket::get_id() {
 }
 
 uint16_t CommsPacket::get_info() {
-	return *reinterpret_cast<uint16_t*>(raw + TEENSY_PACKET_INFO_OFFSET);
+	return *reinterpret_cast<uint8_t*>(raw + TEENSY_PACKET_INFO_OFFSET);
 }
 
 void CommsPacket::set_id(uint16_t id) {
@@ -17,7 +17,7 @@ void CommsPacket::set_id(uint16_t id) {
 }
 
 void CommsPacket::set_info(uint16_t info) {
-	*reinterpret_cast<uint16_t*>(raw + TEENSY_PACKET_INFO_OFFSET) = info;
+	*reinterpret_cast<uint8_t*>(raw + TEENSY_PACKET_INFO_OFFSET) = info;
 }
 
 void CommsPacket::get_target_state(float state[STATE_LEN][3]) {
@@ -27,7 +27,7 @@ void CommsPacket::get_target_state(float state[STATE_LEN][3]) {
 void CommsPacket::get_ref_draw_data(char** draw_data) {}
 
 void CommsPacket::set_time(double time) {
-	*reinterpret_cast<double*>(raw + TEENSY_PACKET_TIME_OFFSET) = time;
+	memcpy(raw + TEENSY_PACKET_TIME_OFFSET, &time, sizeof(double));
 }
 
 void CommsPacket::set_estimated_state(float state[STATE_LEN][3]) {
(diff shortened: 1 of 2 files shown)


src/comms/usb_hid.cpp
```diff
@@ -17,7 +17,7 @@ void CommsPacket::set_id(uint16_t id) {
 }
 
 void CommsPacket::set_info(uint16_t info) {
-	*reinterpret_cast<uint16_t*>(raw + TEENSY_PACKET_INFO_OFFSET) = info;
+	*reinterpret_cast<uint8_t*>(raw + TEENSY_PACKET_INFO_OFFSET) = info;
```
StudentKyle, line 20: this looks like a type mismatch to me (info is `uint16_t`. Is this intentional?

src/comms/usb_hid.cpp
```diff
@@ -27,7 +27,7 @@ void CommsPacket::get_target_state(float state[STATE_LEN][3]) {
 void CommsPacket::get_ref_draw_data(char** draw_data) {}
 
 void CommsPacket::set_time(double time) {
-	*reinterpret_cast<double*>(raw + TEENSY_PACKET_TIME_OFFSET) = time;
+	memcpy(raw + TEENSY_PACKET_TIME_OFFSET, &time, sizeof(double));
```
StudentKyle, line 30: what's this change for?

src/comms/usb_hid.cpp
```diff
@@ -51,46 +51,56 @@ void HIDLayer::ping() {
...
-	int bytes_sent = usb_rawhid_send(m_outgoingPacket.raw, UINT16_MAX);
+	int bytes_sent = usb_rawhid_send(m_outgoingPacket.raw, 0);
 	if (bytes_sent == COMMS_PACKET_SIZE) {
 		// increment total number of packets sent and return success
 		m_packetsSent++;
 		return true;
-	} 
-	else {
+	} else {
+		Serial.println("Comms: failed write");
```
StudentKyle, line 103: will serial printing during execution cause problems (even if this is a failure log?)

src/comms/usb_hid.cpp
```diff
@@ -51,46 +51,56 @@ void HIDLayer::ping() {
 		// attempt to read
 		if (read()) {
 			// if we read, attempt to write
-
-			// set the packet to be written's ID
-			m_outgoingPacket.set_id((uint16_t)m_packetsSent);
```
StudentKyle, line 56: No set id?

src/comms/usb_hid.cpp
```diff
@@ -51,46 +51,56 @@ void HIDLayer::ping() {
...
 
 bool HIDLayer::write() {
+	// verify that the first byte is set to something (0xff)
+	// prevents a weird comms issue where the whole packet is shifted left by one byte if the first byte is not ever set
+	m_outgoingPacket.raw[0] = 0xff;
+
 	// attempt to write a full packet
 	// this has a timeout, which is set to it's max value
-	int bytes_sent = usb_rawhid_send(m_outgoingPacket.raw, UINT16_MAX);
+	int bytes_sent = usb_rawhid_send(m_outgoingPacket.raw, 0);
```
StudentKyle, line 97: a full packet of zero size seems strange

StudentKyle (changes requested):
Off by one changes look consistent with the hive PR, and the printing functions are nice. But there's a couple changes with the get/set info, id, and send functions, what are these meant to do? Just making sure they are intentional.

src/comms/usb_hid.cpp
Pandabear1125: Nah thats fine.

src/comms/usb_hid.cpp
Pandabear1125: Nah the set ID should happen manually, so it doesnt mess up custom packets

src/comms/usb_hid.cpp
Pandabear1125: This only makes it try to send, and if it cant, insta return rather than wait an unknown amount of MILLISECONDS lol

src/comms/usb_hid.cpp
Pandabear1125: The reinterpret_cast was crashing for some reason. The memcpy does the same thing

src/comms/usb_hid.cpp
Pandabear1125: Ah, good catch. I'll fix that.

StudentKyle (approved):
Looks good to me, thanks for the clarifications! (I have not tested it though)

StudentKyle:
Tested, all good

PR #41, opened 2024-04-22
Patch encoder finalization

Pandabear1125:


StudentKyle (approved):
Formatting changes look good.

Removing baudrate from RevEncoder also seems good so multiple Serial.begins aren't called but I haven't tested it.

Pandabear1125:
Its tested

PR #42, opened 2024-04-22
Fixes all current warnings

Pandabear1125:


src/sensors/ICM20649.cpp
```diff
@@ -13,13 +13,16 @@ void ICM20649::init(CommunicationProtocol protocol)
     switch (protocol)
     {
     case I2C:
+    {
```
StudentKyle, line 16: brackets not consistent with other PR

Pandabear1125: This fixes a cross initialization. This is intended

StudentKyle: can the brackets be made on the same line? Either work to create scope here but just to keep it consistent with the rest of the formatting you just did

StudentKyle: Unless you think it's valuable to be different here since it is to fix an unclear error

StudentKyle: Then it's fine how you have it, it being different here could actually be helpful

Rishi-H (approved):

PR #47, opened 2024-04-26
Finished Current Sensor

Pandabear1125:
The values it gives should be correct. They change according to power draw/usage, but I'm not sure its centered around the correct "zero".


StudentKyle (approved):
Looks good to me, saw Jackson test it.

PR #49, opened 2024-04-30
git hook fix

Rishi-H:


StudentKyle (approved):
strange it's in clean but build calls clean

PR #50, opened 2024-05-09
Feature sensor comms

Pandabear1125:
Adds lidar and ref data over comms. Sister hive PR: https://github.com/CU-Robotics/hive/pull/26


Sammyboyo (approved):
could be good

PR #52, opened 2024-05-18
Feature configuration

Rishi-H:
This firmware has proven to work on multiple robots with config packets being correctly read.


Sammyboyo (approved):
looks good

PR #53, opened 2024-05-22
fixed freakouts, teensy loop was running too fast

Rishi-H:
This fixed a freakout that was caused by our loops not running at 1 ms, but rather at 1/5 ms.


Sammyboyo (approved):
looks good

PR #60, opened 2024-08-05
Adds install script for arduino/teensy to be ran on jetson

cu-boulder-robotics:


cu-boulder-robotics:
Tested on my (jackson) laptop, and jetson, and Kyle's

StudentKyle (approved):
Tested on multiple machines, works

PR #61, opened 2024-08-05
Adds install script to add tytools

cu-boulder-robotics:


StudentKyle (approved):
Good after made changes

PR #65, opened 2024-08-25
Feature debug

Pandabear1125:
# This adds quite a bit
 - Adds Dual Serial support, along side normal RawHID support
 - Adds TeensyDebug library
   - Includes support for gdb and in-code debugging
 - Updates Makefile to be more readable
 - Updated the .gitignore to include what it should've included a year ago (whoops)
   - Removes the library files from being included
 - Added scripts to automate monitoring and gdb starting


Pandabear1125:
Need to test on robot, this introduces very critical changes

Pandabear1125:
> Need to test on robot, this introduces very critical changes

Tested, hid comms remain functional, there is a patch required for Hive to recognize Teensy's HID signature though.

teensy4/usb_desc.h
Rishi-H, line 1: why does rawhid transfer and receive go down to 64?

tools/monitor.sh
Rishi-H, line 1: Did we not have this before? What runs on tycmd monitor? or are you just making a firmware tools dir so we don't have to use tycmd?

Pandabear1125: We did have this, however since we are switching to true Serial, tycmd monitor kinda freaks out, especially since we are now running 2 Serial lines. One line is entirely for GDB support (SerialUSB1), and the other is normal Serial (Serial). This program is needed to differentiate between the two and automatically monitor the normal Serial line, rather than SerialUSB1.

tools/monitor.sh
```diff
@@ -0,0 +1,40 @@
+# path to the /dev/ serial device that the Teensy is connected to
+tty_path=""
+
+# Completely disable ctrl-C since it breaks the monitor
+trap '' INT;
```
StudentKyle, line 5: don't we want it to break when you ctrl-C? This seems like we should avoid ever doing this.

StudentKyle: If it's causing some horrible behavior we definitely don't want, I think there could be a way to fix that over just saying "no canceling allowed"

teensy4/usb_desc.h
StudentKyle: If it's unused it would be better not including it at all so people don't get confused.

src/main.cpp
```diff
@@ -79,8 +81,10 @@ void print_logo() {
 int main() {
     long long loopc = 0; // Loop counter for heartbeat
 
-    Serial.begin(1000000); // the serial monitor is actually always active (for debug use Serial.println & tycmd)
-    print_logo();
+    Serial.begin(115200); // the serial monitor is actually always active (for debug use Serial.println & tycmd)
+	debug.begin(SerialUSB1);
+
+	print_logo();
```
StudentKyle, line 87: strange indentation. Is that intended?

tools/monitor.sh
StudentKyle: also have to make sure we're not leaving a zombie monitor process by doing ctrl D instead.

StudentKyle (changes requested):
duplicate code between this repo and hive, need to choose one repo and keep it there so we don't have to maintain both.

Also see comment about trapping ctrl-c

tools/monitor.sh
StudentKyle: You said you verified that there's no zombie processes, so that's good.

Still want to check this myself, and see if I can make ctrl-c work properly as soon as I get the chance. If this is the only thing holding up the approval let me know.

tools/prepare_gdb.sh
StudentKyle, line 1: duplicate code between `monitor.sh` and `prepare_gdb.sh`.

To make it non-duplicate you could have a script called `get_tty_path.sh` and have it stdout the result. Then process them separately as you do at the end, with:

```bash
tty_path=$(./tools/get_tty_path.sh)

... # same code as before to process it
```
( Errors can be captured as well, but probably not needed: [capture stderr](https://stackoverflow.com/questions/12738460/how-to-get-output-of-a-bash-command-in-a-variable) )

I recommend this just to minimize maintainability issues.

src/main.cpp
```diff
@@ -14,6 +14,8 @@
 #include "sensors/d200.hpp"
 #include "sensors/ACS712.hpp"
 
+#include <TeensyDebug.h>
```
StudentKyle, line 17: stray include in main, I don't see it being used here.

StudentKyle (changes requested):
Some maintanability comments and other minor oversights

src/main.cpp
Pandabear1125: The "debug" object comes from this

.gitignore
```diff
@@ -1,11 +1,19 @@
...
+# Ignore the output binaries
+*.elf
+*.hex
+
+# Ignore the libraries
+*.a
+
+# Ignore any tools/ directory artifacts
+tools/*.txt
+tools/*.out
```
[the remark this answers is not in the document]
Pandabear1125: git scraper and gdb preparer

.gitignore
```diff
@@ -1,11 +1,19 @@
-# Ignore build files
-build/
-build/*
```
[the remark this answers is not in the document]
StudentKyle: maybe not a problem if vscode extension doesn't automatically activate without cmake.

Makefile
```diff
@@ -146,5 +150,42 @@ lib_libs:
...
+# # # Utility Targets # # #
+
+help: 
+	@echo "Basic usage: make [target]"
+	@echo "Targets:"
+	@echo "  build:        compiles the source code and links with libraries"
+	@echo "  build_all:    builds the libraries and the source"
+	@echo "  upload:       builds the source and uploads it to the Teensy"
+	@echo "  gdb:          starts GDB and attaches to the firmware running on a connected Teensy"
+	@echo "  build_libs:   builds the external libraries"
```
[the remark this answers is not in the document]
Pandabear1125: That sounds good.

Yes, its a prereq target for build.

src/main.cpp
Pandabear1125: me when tabs

Makefile
```diff
@@ -146,5 +151,45 @@ lib_libs:
...
+	@echo "  monitor: 	   monitors any actively running firmware and displays serial output"
```
StudentKyle, line 164: me when spacing

tools/prepare_gdb.sh
StudentKyle, line 1: (bump) 

duplicate code between `monitor.sh` and `prepare_gdb.sh`.

To make it non-duplicate you could have a script called `get_tty_path.sh` and have it stdout the result. Then process them separately as you do at the end, with:

```bash
tty_path=$(./tools/get_tty_path.sh)

... # same code as before to process it
```
( Errors can be captured as well, but probably not needed: [capture stderr](https://stackoverflow.com/questions/12738460/how-to-get-output-of-a-bash-command-in-a-variable) )

I recommend this just to minimize maintainability issues.

StudentKyle (changes requested):
me when spacing

StudentKyle:
Also looks like there's a `firmware.hex` file. Likely a remnant from before it was added to .gitignore

tools/prepare_gdb.sh
StudentKyle: @Pandabear1125 
Took another look at this. Here's my implementation (tested functionality of this script, but not with teensy):
`tools/get_tty_path.sh`
``` bash
#!/bin/bash

# This finds the correct ttyACM device. First argument passed in should be the device ID.
# Example usage: tty_path=$(./tools/get_tty_path.sh *-f00)

# path to the /dev/ serial device that the Teensy is connected to
tty_path=""

if [ ! -n "$1" ]; then
    # 1>&2 redirects stdout (1) into stderr (2)
    echo "\"$1\" is not a valid tty ID" 1>&2
    exit 1
fi

# Find the correct serial device
for i in /dev/serial/by-id/*; do
    # compare with first argument passed to this command
	if [[ "$i" == $1 ]]; then
		tty_path=$i
		break
	fi
done

# found the id, now we need the actual path to the specific ttyACM device
if [ -n "$tty_path" ]; then
	tty_path=$(readlink -f $tty_path)
fi

# if not valid
if [ ! -n "$tty_path" ]; then
    # 1>&2 redirects stdout (1) into stderr (2)
    echo "Failed to detect tty_path with id \"$1\"" 1>&2
    echo "This probably means the Teensy is not connected or the Teensy is in bootloader mode (red LED on/blinking)" 1>&2

    exit 1
fi

echo $tty_path # outputs to stdout by default

exit 0
```

StudentKyle: Now that I see it it's OK, but might make some of the existing comments more confusing in the gdb and monitor files when describing the ID. 

So up to you if you want this, but I wanted to give it a shot just to learn (will do the nmcli ethernet fix next, this was practice)

Pandabear1125: Could you explain what you find confusing in the gdb and monitor files regarding the ID?

StudentKyle (approved):
See comment https://github.com/CU-Robotics/firmware/pull/65#discussion_r1744330393

Adding that is up to you like I said, just wanted to give it a try. Since it's only that and the ctrl-c issue I'll approve this. 

If I get time later in the year I might revisit the ctrl-c issue, but your warning about it at the start of monitoring seems sufficient (as long as you have time to read it?). We should make a github issue for it.

Still need to test the debugging and new scripts on a robot/teensy once someone gets a chance, so **do not merge** until someone else or me tests it.

tools/prepare_gdb.sh
StudentKyle: I think the comments are very good. I meant it might be slightly less clear if it reads:

`prepare_gdb.sh`
```bash
# The correct serial device has the following name
# usb-Teensyduino_USB_Custom_*-if02
# Where * is a number unique to that serial device, we dont care about that though
# The important part is the "if02" at the end, its interface number 02
# SerialUSB1 is what interface 02 references

tty_path=$(./tools/get_tty_path.sh *-if02)
```

`monitor.sh`
```bash
# The correct serial device has the following name
# usb-Teensyduino_USB_Custom_*-if00
# where * is a number
# The important part is the "if00" at the end, its interface number 00

tty_path=$(./tools/get_tty_path.sh *-if00)
```

Actually maybe that is better. Maybe would add a comment saying (in the get_tty_path.sh I put above)

```bash
# Find the correct serial device.
# The important part is the ID/interface number at the end, which is compared here
for i in /dev/serial/by-id/*; do
    ...
fi
```

StudentKyle: Also one note is I'm not sure if the script comparison should be 

```bash
[[ "$i" == $1 ]]
```
or
```bash
[[ "$i" == "$1" ]]
```

Couldn't really test it without a teensy, so check that syntax.

Pandabear1125: I tested it, it works. Feel free to add it

Pandabear1125:
Tested on infantry, merging

PR #66, opened 2024-09-03
Fixes rev encoder read causing infinite loop cause it doesnt exist lol

Pandabear1125:
Bruh


Sammyboyo (approved):
Probably good. Worked in one test without issues.

PR #69, opened 2024-09-09
Patch inheritance fix

Pandabear1125:
Fixes some virtual function declarations that cause linker errors on lower optimization levels (<-O1)


Pandabear1125:
Tested on infantry, works

Rishi-H (approved):

PR #70, opened 2024-09-11
updated tytools install to use sudo for curl

ashleycody345:
tytools install script provides errors on some systems if `sudo curl` is not used; corrected


Pandabear1125 (approved):
cool!

PR #73, opened 2024-09-20
Move tty id finding to get_tty_path.sh

StudentKyle:
This reduces duplicate code in `monitor.sh` and `prepare_gdb.sh`, improving maintainability.
Code nearly identical to what was discussed in PR #65. 

I'm not able to test the success case without a teensy, but the failure cases have been tested at least. Should verify it works with a teensy before merging.


Pandabear1125 (approved):
Looks good, works on robot

PR #77, opened 2024-09-29
Patch arm compiler

Pandabear1125:
Adds on to the install_arduino.sh script to auto-install the correct compiler if you're running this on aarch64 (the robots).


Pandabear1125:
Tested on Infantry and Sentry, works.

StudentKyle (approved):
Worked. Looks good.

PR #80, opened 2024-10-09
Feature improved makefile

guywithhat99:
Will now cache files to speed up compile times. Also will run multi threaded for speed purposes. Wow so fast!!!


Pandabear1125 (approved):
Looks great, good job :+1:

PR #85, opened 2024-10-18
Feature controls rework

Sammyboyo:
Changing controls to use any state inputs and output to any number of motors.


src/controls/estimator_manager.hpp
```diff
@@ -19,7 +19,6 @@
 
 #define NUM_IMU_CALIBRATION 50000
 
-
 // Rev encoder pins
```
Pandabear1125, line 22: Should this be moved into the rev encoder's driver files?

teensy4/HardwareSerial1.cpp
Pandabear1125, line 1: We should revert the Serial buffer size increases

Pandabear1125 (changes requested):

src/controls/estimator_manager.hpp
Rishi-H: nah, I don't think so, since we init the encoders in estimator manager.

Pandabear1125 (approved):
appears good

PR #91, opened 2024-11-11
Feature- SD config storage

ashleycody345:
Altered config layer configure procedure to include fully automated storage and loading of configuration packets sent from hive. 
This utility is also able to verify configurations based on robot type, based on configuration packets sent from hive and ref system robot ID. 

This feature went through the following tests:
- Without SD card: robot is able to boot correctly without needing SD card. Robot should still be able to verify that a config packet is correct using ref system (incorrect packet should cause infinite loop with error message).
- With SD card: wrong config from hive should cause infinite loop with error message (currently requiring manual restart) instead of loading (wrong) config like it normally would. When the ref system has a different ID from what is on the SD card packet file (`config.pack`) it should error out and request a packet from hive. If `robot_id.cfg` on the robot is changed, the robot will still configure correctly as long as the ref system robot ID matches that of the stored configuration. The robot should still configure a configuration the exact same whether the robot ID is red team or blue team (difference of 100 in robot ID on ref).
- Checksum: if any inconsistency is found whatsoever between stored data and expected output, a checksum (summing over all of the bytes interpreted as unsigned 8-bit integers) should report this inconsistency to detect corrupted data. If the checksum function is tampered with at all, it should not read an old config file made before the changes. If bytes at any location on the `config.pack` file is changed, the configuration should fail and a new config packet should be requested from hive. If any of the places where the checksum is copied to (before each of the config packets) on the config file are altered, the configuration should fail and a new config packet should be requested from hive.


src/comms/config_layer.cpp
```diff
@@ -1,21 +1,74 @@
 #include "config_layer.hpp"
 
 const Config* const ConfigLayer::configure(HIDLayer* comms) {
+    // attempt SD card load
+    config_SD_init(comms);
+    if (configured) return &config;
+
+    // if no config on SD, then await transmission
     // grab and process all config packets until finished
+    double prev_time = millis();
```
Pandabear1125, line 10: These should be uint32_t types

src/comms/config_layer.cpp
```diff
@@ -1,21 +1,74 @@
...
 
+    // verify that config received matches ref system: if not, error out
+#ifndef CONFIG_OFF_ROBOT
+    Serial.printf("Received robot ID from config: %d\nRobot ID from ref system: %d\n", (int)config.robot_id, ref.ref_data.robot_performance.robot_ID);
+    if ((ref.ref_data.robot_performance.robot_ID % 100) != (int)config.robot_id) {
+        Serial.printf("ERROR: IDs do not match!! Check robot_id.cfg and robot settings from ref system!\n");
+        if (!CONFIG_ERR_HANDLER(CONFIG_ID_MISMATCH)) {
+            // honestly not sure what to put here. in current implementation ERR_HANDLER just while(1) loops
+            // this recursive call probably won't work? but it also shouldn't run? will change later if needed
+            return configure(comms);
```
Pandabear1125, line 35: The recursion in this function is unneeded, at least now. If CONFIG_ERR_HANDLER ends up always returning 0, and the config packet constantly is the wrong ID, this will stack overflow and crash. This is currently impossible but I dont see the need for the recursion.

src/comms/config_layer.cpp
```diff
@@ -1,21 +1,74 @@
...
-    char *out_raw = out->raw;
+void ConfigLayer::config_SD_init(HIDLayer* comms) {
+    // if on robot, we need to wait for ref to send robot_id
+    Serial.println("Waiting for ref system to initialize...");
+    while (ref.ref_data.robot_performance.robot_ID == 0) ref.read();
+    Serial.println("Ref system online");
+
+
+    // check SD
+    if (sdcard.exists("/config.pack")) {
```
Pandabear1125, line 56: put this filepath as a variable (const char*) somewhere and reference the variable.

src/comms/config_layer.cpp
```diff
@@ -184,3 +240,173 @@ void Config::fill_data(CommsPacket packets[MAX_CONFIG_PACKETS], uint8_t sizes[MA
         }
     }
 }
+
+bool ConfigLayer::sd_load() {
+    // num of total bytes in config_packets
+    int config_byte_size = MAX_CONFIG_PACKETS * sizeof(CommsPacket);
```
Pandabear1125, line 246: This should be const

src/comms/config_layer.cpp
```diff
@@ -184,3 +240,173 @@ void Config::fill_data(CommsPacket packets[MAX_CONFIG_PACKETS], uint8_t sizes[MA
...
+#ifdef CONFIG_LAYER_DEBUG
+    Serial.printf("sd_load: computing checksum for stored config\n");
+#endif
+    // need combined checksum of config_bytes, subsec_sizes
+    if ((sd_checksum64(config_bytes, config_byte_size) + sd_checksum64(subsec_sizes, MAX_CONFIG_PACKETS)) != checksum) {     
+        Serial.printf("Checksum for config file does not match stored config, requesting config from hive...\n");
+        return false;
+    }
+
+    Config temp_config;
```
Pandabear1125, line 270: Doccument your reason for the temp_config. You explained it to me, but explain it to everyone else reading.

src/comms/config_layer.cpp
```diff
@@ -184,3 +240,173 @@ void Config::fill_data(CommsPacket packets[MAX_CONFIG_PACKETS], uint8_t sizes[MA
...
+    if (sdcard.touch("/config.pack") != 0) {
+        if (!CONFIG_ERR_HANDLER(CONFIG_TOUCH_FAIL)) return false;
+    }
+    if (sdcard.open("/config.pack") != 0) {
+        if (!CONFIG_ERR_HANDLER(CONFIG_OPEN_FAIL)) return false;
+    }
+
+    // total size of /config.pack: MAX_CONFIG_PACKETS * (sizeof(CommsPacket) + 1)
+    // num of packets * size of each packet == num of bytes for all packets
+    int config_byte_size = MAX_CONFIG_PACKETS * sizeof(CommsPacket);
```
Pandabear1125, line 334: Const

src/comms/config_layer.cpp
```diff
@@ -184,3 +240,173 @@ void Config::fill_data(CommsPacket packets[MAX_CONFIG_PACKETS], uint8_t sizes[MA
...
+
+    case CONFIG_OPEN_FAIL:
+        Serial.println("CONFIG_ERROR::config failed to open /config.pack from SD");
+        return false;
+
+    case CONFIG_ID_MISMATCH:
+        prev_time = (float)millis()/1000;
+        delta_time = (float)millis()/1000 - prev_time;
+        while (1) {
+            if (delta_time >= 2.0f) {
```
Pandabear1125, line 397: millis() returns a uint32_t. Just use the integer value. No need to convert to float

Pandabear1125 (changes requested):
Untested by me. But code looks good. A few changes requested but great work!

src/comms/config_layer.cpp
ashleycody345: I converted it to float so that I could measure the time in seconds. However, I realize that I can simply measure in milliseconds to achieve the same thing- so I will go ahead and change that. (Just wanted to explain briefly why I did it so I don't look fully insane)

src/comms/config_layer.cpp
```diff
@@ -1,21 +1,80 @@
...
+    else Serial.println("Config not successfully stored (is an SD card inserted?)");    // not a fatal error, can still return
+
     return &config;
 }
 
-void ConfigLayer::process(CommsPacket *in, CommsPacket *out) {
-    char *in_raw = in->raw;
-    char *out_raw = out->raw;
+void ConfigLayer::config_SD_init(HIDLayer* comms) {
+    const char* config_path = "/config.pack";
```
Pandabear1125, line 53: It would probably be better to store this in the header so you dont need to declare it everywhere you need it.

Pandabear1125 (changes requested):
On top of this, I'd like you to describe the .pack file data structure briefly in the header.

src/comms/config_layer.cpp
ashleycody345: By this, do you mean to simply declare a macro for it (`#include CONFIG_PATH "/config.pack"`)?

Pandabear1125: Yeah, that or something like `const char* CONFIG_PATH = "blah.pack"`. I just want it to be easily changeable.

Pandabear1125: Looks good though

Pandabear1125 (approved):
Requires testing, but code looks good and appears functionally unchanged since I saw it tested.

Pandabear1125:
Dont merge until all reviewers have approved

ashleycody345:
committed to the wrong branch- how embarrassing...!

PR #96, opened 2024-11-30
Feature can refactor

Pandabear1125:
Fully refactors our CAN system. This refactor allows:
 - Fully dynamic motor creation through configuration data 
 - A simplified interface for writing to motors
 - Implements CAN bus 3
 - Abstracts the interface away from motors and into individual motor controllers

**Not in the scope of this PR**
 - Integration with controls
 - Accessing motors via name


ashleycody345 (commented):
I wasn't explicitly requested for an approving review, so I am simply leaving comments with my thoughts. Anything I did not comment on explicitly was something that I thought was intuitive enough to not warrant any changes. Overall I think this looks great!

src/sensors/can/C620.cpp
```diff
@@ -0,0 +1,81 @@
...
+    // set ID
+    msg.id = m_output.id;
+
+    // set only the buffer bytes that correspond to this motor
+    // get the per-struct motor ID
+    uint8_t motor_id = (m_id - 1) % 4;
+
+    // fill in the output array
+    msg.buf[motor_id * 2] = m_output.buf[motor_id * 2];         // upper byte
+    msg.buf[motor_id * 2 + 1] = m_output.buf[motor_id * 2 + 1]; // lower byte
```
ashleycody345, line 37: again, description of buffer data would be helpful (apologies for mentioning this *every time*, but I want to just note all the places where I feel it can be more clear)

src/sensors/can/C620.cpp
```diff
@@ -0,0 +1,81 @@
...
+        m_output.id = 0x200;    // first 4 motors
+    else
+        m_output.id = 0x1FF;    // last 4 motors
+
+    // get the per-struct motor ID
+    uint8_t motor_id = (m_id - 1) % 4;
+
+    // fill in the output array
+    m_output.buf[motor_id * 2] = (int_torque >> 8) & 0xFF;  // upper byte
+    m_output.buf[motor_id * 2 + 1] = int_torque & 0xFF;     // lower byte
```
ashleycody345, line 72: same as before, buffer description would be useful

src/sensors/can/MG8016EI6.cpp
```diff
@@ -0,0 +1,548 @@
...
+        break;
+    }
+    case CMD_READ_ACCELERATION: {
+        m_acceleration = (int32_t)((msg.buf[7] << 24) | (msg.buf[6] << 16) | (msg.buf[5] << 8) | msg.buf[4]);
+        break;
+    }
+    case CMD_READ_ENCODER: {
+        m_encoder = (uint16_t)((msg.buf[3] << 8) | msg.buf[2]);
+        m_encoder_raw = (uint16_t)((msg.buf[5] << 8) | msg.buf[4]);
+        m_encoder_offset = (uint16_t)((msg.buf[7] << 8) | msg.buf[6]);
```
ashleycody345, line 74: It's a little more obvious to me how all of these buffer operations work- I would still request that you make sure the message buffer structure is described somewhere.

ashleycody345 (approved):
Looks great! I would ask that you wait for one other person to approve before merging though

src/sensors/can/can_manager.hpp
```diff
@@ -0,0 +1,108 @@
...
+    void init_motors();
+
+    /// @brief Iterates through all the motors and tries to give the message to the correct one
+    /// @param msg The message to distribute
+    /// @return True if the message was distributed, false if it was not
+    bool distribute_msg(CAN_message_t& msg);
+
+private:
+    /// @brief CAN bus 1
+    FlexCAN_T4<CAN1, RX_SIZE_256, TX_SIZE_16> m_can1;
```
StudentKyle, line 91: Can you add a comment here about things breaking if this is constructed again? When this happens, why this happens (even if that's just flexcan weird library issues)

Also consider static keyword

src/sensors/can/can_manager.cpp
```diff
@@ -0,0 +1,262 @@
...
+        
+        // write the combined messages to the bus
+        m_busses[bus]->write(rm_motor_msgs[0]);
+        m_busses[bus]->write(rm_motor_msgs[1]);
+    }
+}
+
+void CANManager::safety_mode() {
+    // for each motor, cant be const
+    for (auto& motor : m_motor_map) {
```
StudentKyle, line 153: explicit type here would be easier to read

src/sensors/can/motor.hpp
```diff
@@ -0,0 +1,118 @@
...
+    /// @return The bus ID
+    inline uint32_t get_bus_id() const { return m_bus_id; }
+
+    /// @brief Get the current state of the motor
+    /// @return The current state of the motor
+    inline MotorState get_state() const { return m_state; }
+
+protected:
+    /// @brief What controller this motor uses
+    MotorControllerType m_controller_type = NULL_MOTOR_CONTROLLER_TYPE;
```
StudentKyle, line 96: discussed that this information is here in case we want to downcast to a motor with more specialized behavior

src/sensors/can/motor.hpp
```diff
@@ -0,0 +1,118 @@
...
+public:
+    /// @brief Deleted default constructor. Must explicitly construct this object
+    Motor() = delete;
+
+    /// @brief Main constructor. Defines the motor and controller type, global ID, id, and can bus
+    /// @param controller_type The motor controller type 
+    /// @param gid The global ID, not the per-bus motor ID
+    /// @param id The per-bus motor ID. This is 1-indexed 
+    /// @param bus The CAN bus index/ID
+    Motor(MotorControllerType controller_type, uint32_t gid, uint32_t id, uint8_t bus)
```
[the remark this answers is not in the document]
StudentKyle: Answered in discussion: 
CANManager would break due to FlexCan member variables breaking with duplicate constructions (fun).

So, would be good with CANManager having restricted construction and copies. Deleting the default constructor for Motors is fine since they would be in a broken state if forgotten to explictly construct, but Motors should be fine to copy with =.

src/sensors/can/C610.hpp
```diff
@@ -0,0 +1,75 @@
+#ifndef C610_DRIVER_HPP
+#define C610_DRIVER_HPP
+
+#include "motor.hpp"
+
+// C610 Data Sheet
+// https://rm-static.djicdn.com/tem/17348/RoboMaster%20C610%20Brushless%20DC%20Motor%20Speed%20Controller%20User%20Guide.pdf
+
+/// @brief C610 controller driver. This manages generating CAN output messages and processing incomming CAN messages into a state array.
+/// @note It's construction is heavily managed since copying this object could alter the actions of CAN (and by extension the robot). This object exists only to be managed by CANManager.
```
StudentKyle, line 10: see other review comment, allow motors to be copied. Explain the issue with CANManager (which is not an issue with motors)

src/sensors/can/C620.hpp
```diff
@@ -0,0 +1,75 @@
+#ifndef C620_DRIVER_HPP
+#define C620_DRIVER_HPP
+
+#include "motor.hpp"
+
+// C620 Data Sheet
+// https://rm-static.djicdn.com/tem/17348/RoboMaster%20C620%20Brushless%20DC%20Motor%20Speed%20Controller%20V1.01.pdf
+
+/// @brief C620 controller driver. This manages generating CAN output messages and processing incomming CAN messages into a state array.
+/// @note It's construction is heavily managed since copying this object could alter the actions of CAN (and by extension the robot). This object exists only to be managed by CANManager.
```
StudentKyle, line 10: see other review comment, allow motors to be copied. Explain the issue with CANManager (which is not an issue with motors)

src/sensors/can/MG8016EI6.hpp
```diff
@@ -0,0 +1,341 @@
+#ifndef MG8016EI6_DRIVER_HPP
+#define MG8016EI6_DRIVER_HPP
+
+#include "motor.hpp"
+
+// MG8016E-i6 motor data sheet
+// http://en.lkmotor.cn/upload/20230321150047f.pdf
+
+// MG8016E-i6 CAN protocol data sheet
+// http://en.lkmotor.cn/upload/20230706100134f.pdf
```
StudentKyle, line 10: put document in class's doxygen brief

src/sensors/can/can_manager.hpp
```diff
@@ -0,0 +1,108 @@
...
+
+/// @brief CAN bus 1 ID
+constexpr uint32_t CAN_1 = 0u;
+/// @brief CAN bus 2 ID
+constexpr uint32_t CAN_2 = 1u;
+/// @brief CAN bus 3 ID
+constexpr uint32_t CAN_3 = 2u;
+
+// Because of how volatile FlexCAN_T4 objects are, there must only ever exist one instance of this class
+// I'm deleting any copy and assignment constructors that may act to create a new instance
```
StudentKyle, line 22: include this comment somewhere in docs (in CANManager brief and on Flexcan variables)

src/sensors/can/can_manager.cpp
```diff
@@ -0,0 +1,262 @@
...
+        CAN_message_t rm_motor_msgs[2];
+
+        // for each motor, can be const
+        for (const auto& motor : m_motor_map) {
+            // based on the motor type, figure out how to write the message
+            switch (motor.second->get_controller_type()) {
+            case C610_CONTROLLER:   // fallthrough
+            case C620_CONTROLLER: {
+                // depending on the motor ID, write the message to the correct msg in the array
+                if ((motor.second->get_id() - 1) / 4) {
```
StudentKyle, line 116: why these magic numbers on id?

src/sensors/can/C610.cpp
```diff
@@ -0,0 +1,84 @@
...
+    // get the per-struct motor ID
+    uint8_t motor_id = (m_id - 1) % 4;
+
+    // fill in the output array
+    // message format is specified in the C610 datasheet (Speed Controller Receiving Message Format)
+    msg.buf[motor_id * 2] = m_output.buf[motor_id * 2];         // upper byte
+    msg.buf[motor_id * 2 + 1] = m_output.buf[motor_id * 2 + 1]; // lower byte
+
+    // return where in the buffer array the motor data was written
+    return motor_id * 2;
```
StudentKyle, line 42: why multiplied by 2?

libraries/FlexCAN_T4/FlexCAN_T4.h
```diff
@@ -327,6 +327,7 @@ class FlexCAN_T4_Base {
     virtual uint64_t events() = 0;
     virtual int write(const CANFD_message_t &msg) = 0;
     virtual int write(const CAN_message_t &msg) = 0;
+    virtual int read(CAN_message_t &msg) = 0;
```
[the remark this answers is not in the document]
StudentKyle: Discussed: library maybe just forgot to add it?

StudentKyle (changes requested):

src/sensors/can/can_manager.cpp
```diff
@@ -0,0 +1,262 @@
...
+    // for each bus
+    for (uint32_t bus = 0; bus < CAN_NUM_BUSSES; bus++) {
+        // the c610s and c620s require combined messages so are treated differently
+        // I refer to them as rm motors as they are from RoboMaster
+        // first msg is for the first 4 motors, second msg is for the last 4 motors
+        // both the c610s and c620s can occupy the same message
+        CAN_message_t rm_motor_msgs[2];
+
+        // for each motor, can be const
+        for (const auto& motor : m_motor_map) {
```
StudentKyle, line 110: would rather not use auto

src/sensors/can/can_manager.cpp
```diff
@@ -0,0 +1,263 @@
...
+            }
+            default: {
+                Serial.printf("CANManager tried to write to a motor of invalid type: %d\n", motor.second->get_controller_type());
+                break;
+            }
+            }
+        }
+        
+        // write the combined messages to the bus
+        m_busses[bus]->write(rm_motor_msgs[0]);
```
Rishi-H, line 147: is it safe to write default constructed CAN_message_t objects? What is the behavior if either or both of the rm_motor_msgs don't get set with a c610/20 write?

Rishi-H: Okay I looked at previous safety code and it looks like sending a buffer of 0's is how you did safety mode so this should be fine.

src/sensors/can/can_manager.cpp
```diff
@@ -0,0 +1,263 @@
...
+    // using the motor_info array, create the motors following the config
+    
+    // loop through all CAN_MAX_MOTORS
+    for (uint32_t motor_id = 0; motor_id < CAN_MAX_MOTORS; motor_id++) {
+        // grab the information for this specific motor
+        int controller_type = (int)motor_info[motor_id][0];
+        int physical_id = (int)motor_info[motor_id][1];
+        int bus_id = (int)motor_info[motor_id][2];
+
+        // if the controller type is 0, then this motor is unused and we should process the next one
```
Rishi-H, line 45: Looks like this differs from the yaml on hive side, where the controllers are 0 indexed, whereas you use 1 index. I think this code would make it impossible to initialize C610's, as they have a type of 0 in the yaml. I believe you can update the MotorControllerType to have NULL_MOTOR_CONTROLLER_TYPE = -1 and it'll update the rest of the controller ID's correctly. Also, you can then use the NULL_MOTOR_CONTROLLER_TYPE to check if the motor is unused to make it a bit more readable, but I don't really care about that as much.

src/sensors/can/can_manager.cpp
```diff
@@ -0,0 +1,263 @@
...
+            break;
+        }
+        default: {
+            Serial.printf("CANManager tried to create a motor of invalid type: %d\n", controller_type);
+            break;
+        }
+        }
+
+        // place the new motor in the map
+        m_motor_map.insert({ motor_id, new_motor });
```
Rishi-H, line 77: Looks like it's possible to add a nullptr motor object into the map, not sure if you check for that in other places but could be good to add a check here.

src/sensors/can/can_manager.cpp
```diff
@@ -0,0 +1,263 @@
...
+            }
+        }
+    }
+
+    // print out any motors that failed to initialize using the motors_initialized array
+    // this is not fatal but should be investigated
+    for (uint32_t motor = 0; motor < CAN_MAX_MOTORS; motor++) {
+        // if this motor is not initialized and it actually exists, print a warning
+        if (!motors_initialized[motor] && m_motor_map.count(motor) != 0) {
+            Serial.printf("Motor %d on bus %d with id %d failed to initialize\n", motor, m_motor_map[motor]->get_bus_id(), m_motor_map[motor]->get_id());
```
Rishi-H, line 248: It'd be helpful to print the timeout here if a motor fails to initialize

src/sensors/can/can_manager.cpp
```diff
@@ -0,0 +1,263 @@
...
+        }
+    }
+}
+
+bool CANManager::distribute_msg(CAN_message_t& msg) {
+    // for each motor, cant be const
+    for (auto& motor : m_motor_map) {
+        // if the motor can handle the message, give it to the motor
+        if (motor.second->read(msg)) {
+            return true;
```
Rishi-H, line 258: Just curious here, is a can message per motor, or can one can message give information to multiple motors?

src/sensors/can/can_manager.hpp
```diff
@@ -0,0 +1,118 @@
...
+    /// @brief Read data from all busses and distribute them to the correct motors
+    void read();
+
+    /// @brief Issue write commands to all motors on each bus
+    /// @note This issues a CAN command over the bus
+    void write();
+
+    /// @brief Issue zero torque commands to all motors
+    /// @note This immediately issues a CAN command over the bus
+    void safety_mode();
```
Rishi-H, line 69: the name safety_mode kinda implies that it's a toggle and not something that you call every loop to write 0 torque to all the motors. Maybe something like issue_safety_command() could be good.

src/sensors/can/can_manager.hpp
```diff
@@ -0,0 +1,118 @@
...
+
+    /// @brief Print the state of a specific motor
+    /// @param motor_gid The global ID of the motor to print the state of
+    void print_motor_state(uint32_t motor_gid);
+
+    /// @brief Get the underlying motor object by ID
+    /// @param motor_gid The global ID of the motor to get
+    /// @return The motor object if it exists, nullptr if it does not
+    /// @note You can use the motor's get_type to determine the type of motor and dynamic_cast to the correct motor type
+    Motor* get_motor(uint32_t motor_gid);
```
Rishi-H, line 88: When do you think this would be called? I feel like it's a bit dangerous to allow somebody to be able to directly issue commands to a motor outside of the can_manager, at the very least maybe add a note saying don't write anything to this motor, just the functions in can_manager, or just return a copy of the motor instead.

Rishi-H (changes requested):
I really like the structure and how everything's set up. I was a bit nitpicky here but overall I think this is super clean and easy to read/use.

src/sensors/can/can_manager.cpp
Pandabear1125: Both, the rm motors run 4 motors per message, but the mg motors (and all other motors) run it 1-to-1

src/sensors/can/can_manager.hpp
Pandabear1125: This just exposes the underlying motor type. The only thing that can actually issue CAN commands is the can manager. The actions that you could do with the underlying motor type would just expose it's special commands (for example the MG motor's pid controls or start/stop).

src/sensors/can/can_manager.cpp
Rishi-H: Wait so then if you return on the first motor that accepts a message, aren't 3 other motors missing out on data?

Rishi-H (approved):
Looks good

src/sensors/can/motor.hpp
```diff
@@ -7,7 +7,7 @@
 
 /// @brief Defines the motor controller types. Enum values chosen based on the yaml config specification
 enum MotorControllerType {
```
StudentKyle, line 9: enum class would make it so you don't have to worry about the implicit type, and check at compile time any comparisons are valid.

StudentKyle (changes requested):

src/sensors/can/C610.hpp
```diff
@@ -0,0 +1,65 @@
+#ifndef C610_DRIVER_HPP
+#define C610_DRIVER_HPP
+
+#include "motor.hpp"
+
+/// @brief C610 controller driver. This manages generating CAN output messages and processing incomming CAN messages into a state array.
+/// @brief Datasheet: https://rm-static.djicdn.com/tem/17348/RoboMaster%20C610%20Brushless%20DC%20Motor%20Speed%20Controller%20User%20Guide.pdf
+/// @note It's construction should be closely managed. This object exists only to be managed by CANManager.
```
StudentKyle, line 8: This comment is no longer necessarily true

src/sensors/can/C620.hpp
```diff
@@ -0,0 +1,65 @@
+#ifndef C620_DRIVER_HPP
+#define C620_DRIVER_HPP
+
+#include "motor.hpp"
+
+/// @brief C620 controller driver. This manages generating CAN output messages and processing incomming CAN messages into a state array.
+/// @brief Datasheet: https://rm-static.djicdn.com/tem/17348/RoboMaster%20C620%20Brushless%20DC%20Motor%20Speed%20Controller%20V1.01.pdf
+/// @note It's construction should be closely managed. This object exists only to be managed by CANManager.
```
StudentKyle, line 8: This comment is no longer necessarily true.

src/sensors/can/MG8016EI6.hpp
```diff
@@ -0,0 +1,329 @@
+#ifndef MG8016EI6_DRIVER_HPP
+#define MG8016EI6_DRIVER_HPP
+
+#include "motor.hpp"
+
+/// @brief MG8016E-i6v3 controller driver. This manages generating CAN output messages and processing incomming CAN messages into a state array.
+/// @brief Motor datasheet: http://en.lkmotor.cn/upload/20230321150047f.pdf
+/// @brief CAN datasheet: http://en.lkmotor.cn/upload/20230706100134f.pdf
+/// @note It's construction should be closely managed. This object exists only to be managed by CANManager.
```
StudentKyle, line 9: This comment is no longer necessarily true

src/sensors/can/can_manager.hpp
```diff
@@ -0,0 +1,118 @@
...
+/// @note The construction of FlexCAN_T4 objects in of themselves alters processsor state. 
+/// When these are constructed incorrectly, it will exibit the folowing behavior: 
+///     reads are fine, 
+///     the first write will not hang, the processor will lock up some small (seemingly random) amount of time after the write call. 
+/// Sometimes this even allows lines of code after the first write call to be ran (like prints). 
+/// Examples of incorrect construction of FlexCAN_T4 objects include: 
+///     construction in function scope (including main), 
+///     construction via new (likely, although this is never tested. It is an active issue on FlexCAN_T4's github though), 
+///     and construction via copy/assignment. 
+/// This is why we heavily manage the construction of the CANManager
```
StudentKyle, line 34: Thanks for the good explanation

StudentKyle (changes requested):
Good explanation on the FlexCAN issue. The comments on the motors no longer apply (specifically, about needing to manage their construction and copies) since they don't rely on flexcan issue at all, but you can still say "Intended for use by CANManager"

StudentKyle (approved):
Code is good! Just need some final testing before we merge.

Pandabear1125:
Tested

Pandabear1125:
Still here due to complications. Branch will be deleted when the integration PR goes through

PR #97, opened 2024-12-27
Patch off robot macro

ashleycody345:
Found and fixed a very small error with the `CONFIG_OFF_ROBOT` macro found in config layer, as well as very slight reformatting to fit our style guidelines


StudentKyle (approved):
Didn't break anything on the robot

PR #98, opened 2025-01-08
Feature hot reload

Pandabear1125:
Implements hot reloading. This enables Teensy to be reconfigured after startup (say if Hive goes down and comes back up and wants to configure Teensy differently). 

This also has the nice side effect of making Hive resilient to Teensy crashes. Teensy can now die and come back to life and Hive still can maintain a connection with it. 

Teensy is also now resilient to Hive crashes, since Hive can go down and come back up, triggering another configure.

### DIFF
--- a/src/comms/config_layer.cpp
+++ b/src/comms/config_layer.cpp
@@ -12,10 +12,12 @@ extern "C" void reset_teensy(void) {
 	while (1);
 }
 
-const Config* const ConfigLayer::configure(HIDLayer* comms) {
-    // attempt SD card load
-    config_SD_init(comms);
-    if (configured) return &config;
+const Config* const ConfigLayer::configure(HIDLayer* comms, bool config_off_SD) {
+    if (config_off_SD) {
+        // attempt SD card load
+        config_SD_init(comms);
+        if (configured) return &config;
+    }
 
     // if no config on SD, then await transmission
     // grab and process all config packets until finished
@@ -37,7 +39,7 @@ const Config* const ConfigLayer::configure(HIDLayer* comms) {
     config.fill_data(config_packets, subsec_sizes);
 
     // verify that config received matches ref system: if not, error out
-#ifndef DISREGARD_REF_CONFIG_CHECK
+#ifndef DISABLE_REF_CONFIG_SAFETY_CHECK
     Serial.printf("Received robot ID from config: %d\nRobot ID from ref system: %d\n", (int)config.robot, ref.ref_data.robot_performance.robot_ID);
     // id check with modulo 100 to account for red and blue teams. Blue is the id + 100. (ID == 101, 102, ...)
     if ((ref.ref_data.robot_performance.robot_ID % 100) != (int)config.robot) {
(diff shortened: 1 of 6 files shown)


src/comms/config_layer.cpp
```diff
@@ -1,5 +1,17 @@
...
+/// @brief This resets the whole processor and kicks it back to program entry (teensy4/startup.c)
+/// @param void void
+/// @note Dont abuse this function, it is not to be used lightly
+extern "C" void reset_teensy(void) {
+    // Register information found in the NXP IM.XRT 1060 reference manual
+	SRC_GPR5 = 0x0BAD00F1;
+    // Register information found in the Arm-v7-m reference manual
+	SCB_AIRCR = 0x05FA0004;
+    // loop go brr
+	while (1);
```
ashleycody345, line 12: I don't think there's any way for us to actually enter this loop, but do we know what happens if we do for any reason? Firmware should just lock up entirely, if I'm not mistaken.

This loop is absolutely how this should be implemented, I'm just curious to know if we have any failsafes if this loop ever does get entered for some reason.

src/main.cpp
```diff
@@ -156,6 +155,16 @@ int main() {
         CommsPacket* incoming = comms.get_incoming_packet();
         CommsPacket* outgoing = comms.get_outgoing_packet();
 
+        // check whether this packet is a config packet
+        if (incoming->raw[3] == 1) {
```
ashleycody345, line 159: This looks like it relies on the current implementation for how HID handles sending/receiving config packets. How much would this need to be edited to include support for Ethernet down the road- would we really only need to change this one line and make sure that we use the appropriate configure procedure?

ashleycody345 (approved):
I had a few minor concerns about how certain functions might work, so I left some comments asking for general information. However, I don't think there is anything wrong with how this is implemented at the moment. Great work!

src/comms/config_layer.cpp
Pandabear1125: As discussed, the system reset call isnt instant. It may take multiple cycles to complete. The loop is to catch the program so it doesnt do anything else in that time it takes to reset.

src/main.cpp
Pandabear1125: This line would only need to be updated to check whether the incoming generic packet is marked as config. However that is done is left to the implemention. Essentially this check is if (incoming is config)

src/comms/config_layer.cpp
```diff
@@ -125,8 +125,8 @@ void ConfigLayer::reconfigure(HIDLayer* comms) {
 
     // update stored config, msg if successful
     Serial.println("Attempting to store config...");
-    if (store_config()) Serial.println("Config successfully stored in /config.pack");
-    else Serial.println("Config not successfully stored (is an SD card inserted?)");    // not a fatal error, can still return
+    if (store_config()) Serial.println("\tConfig successfully stored in /config.pack\n");
+    else Serial.println("\tConfig not successfully stored (is an SD card inserted?)\n");    // not a fatal error, can still return
```
ashleycody345, line 129: This is a `Serial.println` call and not a `Serial.printf` one - is the `\n` necessary?

ashleycody345 (commented):
Looks fine overall, just small questions

src/comms/config_layer.cpp
Pandabear1125: Yeah it makes the output easier to read

Pandabear1125:
Tested on sentry and infantry, no explosions

src/comms/config_layer.cpp
```diff
@@ -1,5 +1,17 @@
 #include "config_layer.hpp"
 
+/// @brief This resets the whole processor and kicks it back to program entry (teensy4/startup.c)
+/// @param void void
```
StudentKyle, line 4: Haven't seen this before, but worth documenting since that is just weird historical C that isn't present in C++.
```suggestion
/// @param void specify no arguments (needed in C)
```

src/comms/config_layer.hpp
```diff
@@ -26,8 +26,8 @@
 // lastly, stores bytes of subsec_sizes array
 #define CONFIG_PATH "/config.pack"
 
-// define CONFIG_OFF_ROBOT macro when running off of real robot (testing firmware away from actual robot)
-// #define CONFIG_OFF_ROBOT 
+// define DISREGARD_REF_CONFIG_CHECK macro when running off of real robot (testing firmware away from actual robot)
+#define DISREGARD_REF_CONFIG_CHECK
```
StudentKyle, line 30: `IGNORE_REF_CONFIG_SAFETY` or something similar with the word `SAFETY` in it sounds scarier, making it clear that disabling it has safety implications. Or `DISABLE` as well to be clearer that this is also disabling ref elsewhere (if I understand correctly).

src/comms/config_layer.cpp
```diff
@@ -42,17 +54,96 @@ const Config* const ConfigLayer::configure(HIDLayer* comms) {
...
+    // Hive marks the config process as done once teensy sends a packet with info_bit == 0
+    // Hive then sends a packet back with info_bit == 0 as well, so ping until we get a non-config back
+    while (comms->get_incoming_packet()->raw[3] == 1) {
+        comms->ping();
+    }
 
     return &config;
 }
 
+void ConfigLayer::reconfigure(HIDLayer* comms) {
```
StudentKyle, line 81: configure and reconfigure can't have this much duplicate code between them for maintainability reasons. I'd say either make everything just call the one `configure` and make them do the same thing, or have reconfigure call the configure function inside of it, so you can have your before and after behavior still there (which seems to be the only difference).

src/comms/config_layer.hpp
```diff
@@ -7,7 +7,7 @@
 
 #include <map>
 #include <string>
-#define CONFIG_LAYER_DEBUG
+// #define CONFIG_LAYER_DEBUG
```
StudentKyle, line 10: was this intentional?

StudentKyle (changes requested):
Some feedback. Excited for this feature!

src/comms/config_layer.hpp
Pandabear1125: no lol

src/comms/config_layer.hpp
StudentKyle, line 1: comms parameter is not documented in config_SD_init

https://github.com/CU-Robotics/firmware/blob/9dad580e276fbd394dab5e5f22d8569043108d60/src/comms/config_layer.hpp#L155

src/comms/config_layer.hpp
```diff
@@ -26,9 +25,8 @@
 // lastly, stores bytes of subsec_sizes array
 #define CONFIG_PATH "/config.pack"
 
-// define CONFIG_OFF_ROBOT macro when running off of real robot (testing firmware away from actual robot)
-// #define CONFIG_OFF_ROBOT 
-
+// define DISABLE_REF_CONFIG_SAFETY_CHECK macro when running off of real robot (testing firmware away from actual robot)
+#define DISABLE_REF_CONFIG_SAFETY_CHECK
```
StudentKyle, line 29: remember to un-define this before merge.

StudentKyle (changes requested):
Looks great! Just noticed an undocumented parameter, and also remember to un-define the disable safety check.

src/comms/config_layer.hpp
Pandabear1125: oop thanks haha

StudentKyle (approved):
Good

StudentKyle:
@Pandabear1125 can add the edge case documentation in another PR, but he says the tests passed, so merging.

PR #101, opened 2025-01-13
Revert "Feature can refactor"

Pandabear1125:
Reverts CU-Robotics/firmware#96


Pandabear1125:
Turns out the can code just existing (never called or created) messes with the old can code. Bruh

PR #103, opened 2025-01-17
Feature wfly transmitter

ericv555:
Driver package adds ET16S class that maps raw transmitter data to usable values. This means -1 to 1 for dial, wheel, and stick, as well as 1,2 and 3 for switches. Disconnect messages are also received from the transmitter and can be accessed with a getter function.


src/sensors/ET16S.cpp
```diff
@@ -0,0 +1,389 @@
+#include "ET16S.hpp"
+
+ET16S::ET16S() { }
+
+//EDITING NOTES::
```
Pandabear1125, line 5: Was this ever addressed?

src/sensors/ET16S.cpp
```diff
@@ -0,0 +1,389 @@
...
+}
+
+void ET16S::read() {
+	// Raw data stored in array
+	uint8_t m_inputRaw[ET16S_PACKET_SIZE] = { 0 };
+	if (Serial8.available() < 50) {
+		return;
+	}
+
+	while (Serial8.peek() != 0x0f) {
```
Pandabear1125, line 42: more general to the whole project, but please comment more. Explain what you do on (almost) every step. Especially if it isnt very obvious how or why youre doing something

src/sensors/ET16S.cpp
```diff
@@ -0,0 +1,389 @@
...
+
+float ET16S::map_raw(InputChannel input) {
+	float max_out = 1;
+	float min_out = -1;
+	float val = input.raw_format;
+	InputKind kind = input.kind;
+
+	switch (kind) {
+	case InputKind::STICK: {
+		val = min_out + (val - min_in) * (max_out - min_out) / (max_in - min_in);
```
Pandabear1125, line 240: make the mapping math its own function. That way you can pass in the mins and maxes. Also, if you generalize it, you can use it for the switches too

src/sensors/ET16S.cpp
```diff
@@ -0,0 +1,389 @@
...
+	}
+	case InputKind::FLAG: {
+		break;
+	}
+	}
+
+	return val;
+}
+
+void ET16S::set_config() {
```
Pandabear1125, line 291: You specify which input kind these channels are physically, but how does anyone know which one is a specific switch? How would i know whether which one is Switch B and the like?

src/sensors/ET16S.hpp
```diff
@@ -0,0 +1,153 @@
+#pragma once
+
+#include <Arduino.h>
+#include <string>
+//http://www.wflysz.com/wflyftp/ET16S/ET16SENV1.00.pdf
+constexpr uint16_t ET16S_PACKET_SIZE = 25;
```
Pandabear1125, line 6: Doxygen wont complain about these, but make sure you do the /// @brief on them please

src/sensors/ET16S.hpp
```diff
@@ -0,0 +1,153 @@
...
+constexpr uint16_t ET16S_INPUT_VALUE_COUNT = 17;
+constexpr float max_in = 1695;
+constexpr float min_in = 352;
+constexpr uint16_t ERROR = 0b0000000000001100;
```
Pandabear1125, line 10: Especially this, document how you know this is an error, even if you just end up saying "found through testing"

src/sensors/ET16S.hpp
```diff
@@ -0,0 +1,153 @@
...
+	/// @return (-1 to 1)
+	float get_r_stick_y();
+
+	/// @brief get left stick y axis value
+	/// @return (-1 to 1)
+	float get_l_stick_y();
+
+	/// @brief get channel 5 data
+	/// @return channel 5 data
+	float get_channel_five();
```
Pandabear1125, line 78: Kyle3 and I were talking earlier today that it might prove better to use on firmware's side if you implement all the getters for each specific input on the controller like dr16 has rather than just specific channels. You can implement it using std::option such that the getter has the ability to return nothing if the channel isn't actually mapped to anything.

src/sensors/ET16S.hpp
```diff
@@ -0,0 +1,153 @@
...
+
+	/// @brief get channel 6 data
+	/// @return channel 6 data
+	float get_channel_six();
```
Pandabear1125, line 82: On top of the eariler comment about the getters, make this just a get_channel(int channel_number) rather than actually specifying the channel name in the actual function name. You dont even need to treat the hard coded channels any different this way

src/sensors/ET16S.hpp
```diff
@@ -0,0 +1,153 @@
...
+	void set_config();
+
+	/// @brief tests flag byte for disconnect, sets safety to zero if disconnect
+	void test_connection();
+
+	/// @brief Array of channels, each holds its respective raw binary data, mapped data, and the kind of input it is (ex: switch,dial,etc...)
+	InputChannel channel[ET16S_INPUT_VALUE_COUNT];
+
+	/// @brief signifies whether a disconnect flag has been read
+	bool is_connected;
```
Pandabear1125, line 152: Initialize this to false

src/sensors/ET16S.hpp
Pandabear1125: One final thing, when you implement the specific getters, specifically for the discrete inputs like switches, make it return an enum. Maybe like UP, MIDDLE, DOWN so it can be more readable when we use it (is 1 up or down?)

src/sensors/ET16S.cpp
```diff
@@ -0,0 +1,389 @@
...
+	}
+
+	for (int i = 0; i < ET16S_PACKET_SIZE;i++) {
+		m_inputRaw[i] = Serial8.read();
+	}
+
+	//format raw data
+	format_raw(m_inputRaw);
+	//set joysticks data
+	channel[0].data = map_raw(channel[0]);
```
Pandabear1125, line 53: If you implement these in the switch statement of set_channel_data, you wouldn't need to handle them separately here.

src/sensors/ET16S.cpp
```diff
@@ -0,0 +1,389 @@
...
+//DIAl and WHEEL are backwards left is highest number
+// right is lowest number
+//Probably flip that
+
+void ET16S::init() {
+	Serial8.begin(100000, SERIAL_8E1_RXINV_TXINV);
+	Serial8.flush();
+	Serial8.clear();
+	//configure flag byte
+	channel[16].kind = InputKind::FLAG;
```
Pandabear1125, line 15: Why not configure these in set_config()?

src/sensors/ET16S.hpp
```diff
@@ -0,0 +1,153 @@
...
+
+	/// @brief Breaks up data packet into correct 11 bit chunks
+	/// @param m_inputRaw takes in 8 bit chunks of the total data packet
+	void format_raw(uint8_t m_inputRaw[ET16S_PACKET_SIZE]);
+
+	/// @brief assigns the mapped input values to their respective channels
+	void set_channel_data();
+
+	/// @brief Defines the kind of input (ex: switch,dial,etc...) for each configurable channel
+	void set_config();
```
Pandabear1125, line 143: This definitely can be addressed after we get this pr merged, but this should take in data from the config and not be set up in-code.

Pandabear1125 (changes requested):
Great work. Some moderate comments about code flow and the desired interface. You did a great job figuring out the data separation and processing. Thanks for working on this!

src/sensors/ET16S.cpp
```diff
@@ -0,0 +1,502 @@
...
+	//right stick horizontal
+	channel[3].kind = InputKind::STICK;
+	//configure remaining channels
+	set_config();
+}
+
+void ET16S::read() {
+	// Raw data stored in array
+	uint8_t m_inputRaw[ET16S_PACKET_SIZE] = { 0 };
+	// Store 2 packets (50 bits) in the buffer to prevent incomplete packet reads
```
Pandabear1125, line 32: This is in bytes. Also make this check using your packet size define rather than just "50"

src/sensors/ET16S.cpp
```diff
@@ -0,0 +1,502 @@
...
+std::optional<float> ET16S::get_l_dial(){
+	if (l_dial_num.has_value()){ return {}; }
+	return channel[l_dial_num.value()].data;
+}
+std::optional<float> ET16S::get_r_dial(){
+	if (r_dial_num.has_value()){ return {}; }
+	return channel[r_dial_num.value()].data;
+}
+
+float ET16S::get_channel_data(int chan_num){
```
Pandabear1125, line 492: What is the reason why I cant access the hard codded channels? This is only a getter so I dont have the ability to remap these. Might as well support it. Also what does -2 mean?

src/sensors/ET16S.hpp
```diff
@@ -0,0 +1,271 @@
...
+	void test_connection();
+
+	/// @brief Array of channels, each holds its respective raw binary data, mapped data, and the kind of input it is (ex: switch,dial,etc...)
+	InputChannel channel[ET16S_INPUT_VALUE_COUNT];
+
+	/// @brief signifies whether a disconnect flag has been read
+	bool is_connected=false;
+	//switch a (safety switch) and joysticks are not configurable
+	/// @brief switch b index
+	std::optional<int> switch_b_num=5;
```
Pandabear1125, line 238: do we want these values initialized to a value? I thought these values are filled by the config.

Pandabear1125 (changes requested):

src/sensors/ET16S.hpp
ericv555: I was thinking these will be filled by the config once the config is setup in a later pr, until then they will be set here.

Pandabear1125 (approved):
LGTM :+1:

PR #106, opened 2025-01-23
Feature better monitor

Pandabear1125:


src/comms/config_layer.hpp
```diff
@@ -26,7 +26,7 @@
 #define CONFIG_PATH "/config.pack"
 
 // define DISABLE_REF_CONFIG_SAFETY_CHECK macro when running off of real robot (testing firmware away from actual robot)
-// #define DISABLE_REF_CONFIG_SAFETY_CHECK 
+#define DISABLE_REF_CONFIG_SAFETY_CHECK
```
ashleycody345, line 29: you should probably comment this out again

tools/monitor.c
```diff
@@ -0,0 +1,215 @@
...
+    if (dev_serial == -1) {
+        printf("%s: Failed to open file: %s\n", argv[0], teensy_dev_path);
+        
+        exit_code = EXIT_FAILURE;
+        goto cleanup;
+    }
+
+    // allocate the in_buffer
+    in_buffer = malloc(read_size);
+    if (in_buffer == NULL) { perror("in_malloc_alloc"); goto cleanup; }
```
ashleycody345, line 63: `cleanup` attempts to free both `in_buffer` and `out_buffer`, but this line runs before `out_buffer` ever has malloc called. Plus, if `in_buffer == NULL`, shouldn't `free(in_buffer)` cause a runtime error?

tools/monitor.c
```diff
@@ -0,0 +1,215 @@
...
+
+/// @brief In serial buffer
+char* in_buffer = NULL;
+/// @brief Out serial buffer
+/// @todo Implement writing sometime, this is much more complicated that it seems
+char* out_buffer = NULL;
+
+/// @brief Base directory for serial devices on Linux
+/// @note No clue why FTS needs it in this format but oh well
+char* serial_directory[ ] = { "/dev/serial/by-id/" };
```
ashleycody345, line 26: This works for Ubuntu machines, I assume that the Linux kernel will handle this the same pretty much regardless of distro? Still, this shouldn't work on any other machines, and I'm not sure we want to discard support for devices other than our own (generally)

ashleycody345: Update: finished reading the monitor shell script, it doesn't seem like it's an issue since we have the backup tycmd monitor.

ashleycody345 (changes requested):
Looks good overall, just some slight comments that you might wanna take a look at first. Commenting out the `DISABLE_REF_CONFIG_SAFETY_CHECK` is something I do want you to do though

ashleycody345 (approved):
LGTM 👍

PR #108, opened 2025-01-24
Patch slow loop freakout

Pandabear1125:
Automatically triggers safety mode if our loop time is exceeded.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -294,9 +294,10 @@ int main() {
         // check whether this was a slow loop or not
 	    float dt = stall_timer.delta();
         if (dt > 0.002) { 
-	    	Serial.printf("Slow loop with dt: %f\n", dt);
             // zero the can bus just in case
 	    	can.zero();
+		
+	    	Serial.printf("Slow loop with dt: %f\n", dt);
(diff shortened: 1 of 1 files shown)


Pandabear1125:
Not merging until tested on all robots this meeting.

Sammyboyo (approved):
Looks good.

src/main.cpp
```diff
@@ -296,8 +289,20 @@ int main() {
         outgoing->set_ref_data(ref_data_raw);
         outgoing->set_estimated_state(temp_state);
 
+        bool is_slow_loop = false;
+
+        // check whether this was a slow loop or not
+	    float dt = stall_timer.delta();
+        if (dt > 0.002) { 
+	    	Serial.printf("Slow loop with dt: %f\n", dt);
```
Rishi-H, line 297: Lil nitpicky but ideally can.zero() is called before the print since we want to stop a potential freakout as soon as possible. Looks very good otherwise, and I saw you test it and it seemed good.

Rishi-H (changes requested):

Rishi-H (approved):
LGTM

PR #118, opened 2025-02-17
Adds nano spec to build flags

Pandabear1125:
Links against newlib.nano instead of the normal standard glibc. This dramatically reduces binary size. 

Works on sentry and infantry.


roccoagain:
tested on my teensy, works beautifully

roccoagain (approved):
one line in the makefile what could go wrong

PR #123, opened 2025-03-09
Feature watchdog

ivca8479:
Added watchdog timer functionality


src/utils/watchdog.hpp
```diff
@@ -0,0 +1,84 @@
...
+    double timeout = 1;
+    /// @brief the wdt is the type to initialize the watchdog and the watchdog you used is selected, here watchdog 1 is selected which is used to reset the CPU
+    WDT_T4<WDT1> wdt;
+    /// @brief  struct for the watchdog configurations used to initialize the configurations of the timer, including the callback, the trigger,  and the timout
+    WDT_timings_t config;
+public:
+    /// @brief default watchdog if no user input is given, in order to start the watchdog, myst start watchdog_start()
+    Watchdog(){
+        this->config.trigger = trigger; //default trigger value = 0.5 second
+        this->config.timeout = trigger; //default timeout value = 1 second
```
Pandabear1125, line 27: This is using the wrong variable

src/utils/watchdog.hpp
```diff
@@ -0,0 +1,84 @@
...
+        this->trigger = trigger;
+        this->config.trigger = trigger;
+        this->config.timeout = timeout;
+        this->config.callback = watchdog_callback;
+    }
+    /// @brief this must be called in order to start the watchdog, after this is called, wathcdog 1 is turned on to keep track of when trigger and timouts should occur based on previous initializations of timer and trigger
+    void watchdog_start(){
+        wdt.begin(config);
+    }
+    /// @brief this stops the watchdog, to stop using it if no longer needed
```
Pandabear1125, line 44: didn't we figure out that this triggers the reset? This should be renamed to reflect what it actually does

src/utils/watchdog.hpp
```diff
@@ -0,0 +1,84 @@
+#pragma once
```
Pandabear1125, line 1: Every function in this Watchdog class need to have the "watchdog_" part of their name removed. It is implied because the function exists within Watchdog.

Pandabear1125 (changes requested):
Looks good, just a bit of naming issues

src/utils/watchdog.hpp
ivca8479: That was the function watchdog_reset() that I had defined, watchdog start, starts the watchdog and then it can't be stopped after, can only be reconfigured. I renamed the watchdog_reset() to just reset in case there is a reason to reset the whole CPU

Pandabear1125 (approved):
Tested on the sentry, works. LGTM

PR #126, opened 2025-03-14
Sends crash report over serial

Pandabear1125:
Simple PR, just checks to see if a crash exists, and prints it forever.


effdotsh (approved):
LGTM

roccoagain (commented):
Ran with and without an intentional crash, both behaved as expected.

PR #130, opened 2025-04-02
tested no prints

Rishi-H:
This pr removes a lot of the status prints that slow down the teensy loop and cause weird controller behaviors.


Pandabear1125 (approved):
LGTM :+1:

PR #135, opened 2025-04-27
Patch - clamp mouse integrator

roccoagain:
Clamps dr16_pos_y to the pitch max and min from robot config, so that we dont accumulate mouse movements out of bounds.


Pandabear1125 (approved):
lgtm :+1:

PR #137, opened 2025-04-30
Patch integrator again

roccoagain:
does the range shift *before* the clamp, so that it works properly when the mouse/controller starts from 0.


Pandabear1125 (approved):
real

PR #140, opened 2025-05-28
Fix infinite configure condition 

Rishi-H:
Switch the or to an and so that it doesn't always wait 500 ms even if we got a 0 request from hive.


Rishi-H:
We are not sure this issue is completely fixed, as it was intermittent to begin with. We will keep this pr open and merge it if we are unable to recreate the issue in the next few days.

PR #146, opened 2025-08-13
Feature better mac install

guywithhat99:
This PR Allows Firmware to be easily installed on MacOS using the same install scripts as Linux. The scripts will check what OS the computer is running and install accordingly. This adds a dependency to firmware of the https://github.com/CU-Robotics/tycmd repo as this holds the tycmd binary for the install_tytools script. There is an argument to be made that this should be in the firmware repo instead but I omitted it for now because its only required for Mac install and I think Jackson didn't like that Idea


.gitignore
```diff
@@ -46,3 +46,6 @@ src/git_info.h
 
 # Ignore Rust tool's build directory
 /tools/waggle-interceptor/target
+
+#ignore the compile_commands.json file
+compile_commands.json
```
roccoagain, line 51: compile_commands.json is already in the .gitignore

PR #154, opened 2025-08-24
Feature clangd project support

guywithhat99:
This one is going to be a doozy. I've gone through and fixed errors that clangd found with our project. Mostly with includes. Some classes were only headers. I broke them out into their own cpp file. Take a look and let me know what you think and how you feel about it. It would be nice to go through on a normal vscode setup too to see in intelsense hates anything. All includes start at src.

### DIFF
--- a/src/utils/bitwise_operations.cpp
+++ b/src/utils/bitwise_operations.cpp
@@ -1,0 +1,7 @@
+#include "bitwise_operations.hpp"
+
+uint16_t combine_bytes(uint8_t high, uint8_t low) {
+  uint16_t result = 0;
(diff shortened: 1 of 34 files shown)


src/utils/bitwise_operations.cpp
```diff
@@ -0,0 +1,7 @@
+#include "bitwise_operations.hpp"
+
+uint16_t combine_bytes(uint8_t high, uint8_t low) {
```
roccoagain, line 3: This is unreferenced and kinda trivial, if you want to clean it up

PR #158, opened 2025-09-03
Oh god a bear

guywithhat99:
allows bear to create a compilation database for firmware by using "make cdb". Also gets rid of some unused includes in main that should have been in the previous pr


roccoagain (approved):
Tested on Mac and Ubuntu with VSCode and Zed, LGTM.

PR #161, opened 2025-09-07
Clangd clenaup

guywithhat99:
This pull request removes the FastLED library, also cleans up some errors from other system files. It also will get rid of the comms mega struct.


roccoagain:
This looks like it was branched from feature-vn100 instead of main, judging by the first couple of commits. Could you please make the same changes starting from main? Or take out the vn100 stuff.

PR #169, opened 2025-10-29
Also send keys in get_transmitter_data()

roccoagain:


src/sensors/dr16.cpp
```diff
@@ -285,5 +284,21 @@ TransmitterData DR16::get_transmitter_data(){
 	transmitter_data.r_stick_y = get_r_stick_y();
 	transmitter_data.wheel = get_wheel();
 	transmitter_data.safety_switch = get_safety_switch();
+	transmitter_data.keys.key.w = keys.w;
```
Pandabear1125, line 287: You can just assign `keys` since its the same struct.

src/sensors/ET16S.cpp
```diff
@@ -618,5 +618,21 @@ TransmitterData ET16S::get_transmitter_data(){
 	transmitter_data.trim_four = get_trim_four();
 	transmitter_data.trim_five = get_trim_five();
 	transmitter_data.trim_six = get_trim_six();
+	transmitter_data.keys.key.w = keys.w;
```
Pandabear1125: does the et16s have keys?

Pandabear1125 (approved):
Looks good, 2 comments

PR #170, opened 2025-11-02
update-firmware-documentation-readme

thetopham:


guywithhat99 (commented):
We already have Doxygen support set up for this repo, so any technical documentation about the code (APIs, functions, build flags, etc.) should live there—kept factual and in sync with the source. If you want to write broader or evolving documentation (like workflow notes or explanations that may change over time), consider adding that to an MD book instead, where it’s easier to iterate and refine.

PR #177, opened 2026-02-22
Preflight check

### LINKED ISSUE #171: Pre-flight check for ET16S dials.
The ET16S has non-centering dials on the back. This means the robot could be set to the maximum spin rate before exiting safety mode. This fix should be similar to how FPV drones make you zero the throttle before arming.

vp13t:
Prevent firmware start without being in safety mode. Prevent leaving safety mode without wheel and shooter being zerod.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -123,35 +123,6 @@ int main() {
         transmitter = new ET16S;
     }
 
-    bool transmitter_safe = false;
-    while (!transmitter_safe) {
-        transmitter->read();
-        
-        //make sure there is no spin
-        if(transmitter->get_wheel() != 0) {
-            Serial.println("Preflight fail: Wheel must be 0");
-            delay(1000);
-            continue;
-        }
-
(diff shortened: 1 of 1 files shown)


src/main.cpp
```diff
@@ -469,18 +465,43 @@ int main() {
...
+                requested_mode = SAFETY;
+            }
+            //make sure flywheels and the feeder are off 
+            else if(transmitter->get_r_switch() != SwitchPos::BACKWARD) {
+                Serial.println("Failed to disengage safety mode. Right switch must be in backward position (shooter off).");
+                requested_mode = SAFETY;
+            }
+        }
+        if (requested_mode != mode) {
+            Serial.printf("Switching mode from %d to %d\n", mode, requested_mode);
```
ericv555: Would be great if this printed "hive, safety, teensy" instead of 0,1,2

vp13t: Updated.

ericv555 (approved):
Looks good

PR #183, opened 2026-04-01
Patched incorrect chassis heading angle calculation

Rishi-H:


src/controls/estimator.cpp
effdotsh: pr only makes a difference because `chasis_angle` is being used as two seperate things. lets address the root and rename the vars to something more meaningful

effdotsh (changes requested):

effdotsh (approved):

PR #187, opened 2026-04-12
Feature unit testing

leBronzo1:
Tests for controls, sensors, utils, and filters.


src/controls/controller.cpp
```diff
@@ -179,22 +180,21 @@ void XDriveController::step(RobotStateMap& reference_map, RobotStateMap& estimat
...
-        float chassis_heading = estimate_map[Cfg::StateName::ChassisHeading].get_position();
 
-        // Convert to motor velocities
-        motor_velocity[1] = output[0] * cos(chassis_heading) + output[1] * sin(chassis_heading) + output[2];
-        motor_velocity[2] = output[0] * sin(chassis_heading) - output[1] * cos(chassis_heading) + output[2];
-        motor_velocity[3] = -output[0] * cos(chassis_heading) - output[1] * sin(chassis_heading) + output[2];
-        motor_velocity[0] = -output[0] * sin(chassis_heading) + output[1] * cos(chassis_heading) + output[2];
+        float chassis_heading = estimate_map[Cfg::StateName::ChassisHeading].get_position();
+        MotorVelocities mv = xdrive_mix(output[0], output[1], output[2], chassis_heading);
+        // Mixer order maps to controller motor indices [1,2,3,0].
```
ericv555: Why don't you adjust xdrive_mix so that the order of the output is in the correct order for the motor_velocity. Then there is no need to do the assignments after this you can assign motor_velocity directly from xdrive_mix

src/controls/controller_math.cpp
```diff
@@ -0,0 +1,27 @@
...
+    const float sine = std::sin(heading);
+    return {{
+        x * cosine + y * sine + rot,
+        x * sine - y * cosine + rot,
+        -x * cosine - y * sine + rot,
+        -x * sine + y * cosine + rot,
+    }};
+}
+
+float clamp1(float value) {
```
ericv555: This should be inlined or defined in the header

src/controls/controller_math.hpp
```diff
@@ -0,0 +1,32 @@
...
+/// @param y Translational velocity input along the reference frame's y axis.
+/// @param rot Rotational contribution added to each motor target, in the same units as x and y.
+/// @param heading Chassis heading relative to the translation reference frame, in radians.
+/// @return Unclamped, unnormalized motor velocity targets in the order documented by MotorVelocities::v.
+MotorVelocities xdrive_mix(float x, float y, float rot, float heading);
+
+/// @brief Clamp a controller output to the inclusive range [-1, 1].
+/// @param value Controller output to clamp.
+/// @return The input value constrained to [-1, 1]; NaN inputs remain NaN.
+float clamp1(float value);
```
ericv555: It would probably be best to remove this and revert the code back to std::clamp. I believe this makes the code less readable as I cannot immediatly tell the bounds without checking this function definition.

src/controls/controller_math.cpp
ericv555: As a note if you were to keep this you would want to inline it or define it in the header.

src/controls/controller_math.cpp
ericv555, line 1: I would get rid of this file. I think it is unneccesary due to its brevity. If these functions need their own file keep them inlined in the header. If they do not then you can put them in the namespace defined at the top of controller.cpp

ericv555 (changes requested):

src/controls/controller.cpp
leBronzo1: Fixed mb Im kind of slow

src/controls/controller_math.hpp
leBronzo1: Reverted back to std::clamp

src/controls/controller_math.cpp
leBronzo1: I just deleted the clamp1 no wories about that

src/controls/controller_math.cpp
leBronzo1: Everything in controller with a namespace

PR #189, opened 2026-04-14
Main refactor

ericv555:
Just a little main refactor to keep things interesting. Moved all Main functionality to member functions of a (hypervisor?) class called HelloRobot. This breaks up what was previously main.cpp into 5 primary functions, which can be individually run and tested.  This reduces our main script to simply initializing the robot and running its main loop. The loop itself is now also broken up into the division I suggested above. The HelloRobot class owns all of the managers and hardware interfaces, except for ref and the comms_layer, which have been globalized. The purpose of this architectural change is to allow for unit tests of all main file functionality.


src/sensors/transmitter/dr16.cpp
```diff
@@ -324,30 +324,27 @@ void DR16::manual_controls(const RobotStateMap& estimated_state_map, RobotStateM
...
-	vtm_pos_x += ref.ref_data.kbm_interaction.mouse_speed_x * 0.05 * delta;
-	vtm_pos_y += ref.ref_data.kbm_interaction.mouse_speed_y * 0.05 * delta;
-
+	// Used for determing whether we are using vtm or usb to dr16
+	auto merge_input = [](auto primary_val, auto secondary_val) {
+		// If primary isn't zero, use it. Else use secondary.
+		return (primary_val != 0) ? primary_val : secondary_val;
+	};
+	
+	pos_x = merge_input(mouse_x,ref.ref_data.kbm_interaction.mouse_speed_x) * 0.05 * delta;
```
roccoagain: I believe this breaks mouse aiming. The old code accumulated mouse movement with `+=`. Aiming now only reflects the mouse's tiny delta each loop.

src/sensors/transmitter/dr16.cpp
```diff
@@ -324,30 +324,27 @@ void DR16::manual_controls(const RobotStateMap& estimated_state_map, RobotStateM
...
+	pos_y = merge_input(mouse_y,ref.ref_data.kbm_interaction.mouse_speed_y) * 0.05 * delta;
+	l_mouse_button = merge_input(l_mouse_button,ref.ref_data.kbm_interaction.button_left);
```
roccoagain: This writes the merged value, which is based on `l_mouse_button`, back into `l_mouse_button`. `merge_input(l_mouse_button, ...)` can keep selecting that previous true value even after the button is released, causing it to latch.

src/hello_robot.hpp
```diff
@@ -0,0 +1,168 @@
...
+#include "utils/timing.hpp"
+#include "utils/watchdog.hpp"
+
+extern "C" void reset_teensy(void);
+
+// Loop constants
+#define LOOP_FREQ 1000
+#define HEARTBEAT_FREQ 2
+#ifdef PROFILER
+Profiler prof;
```
roccoagain: Linker error when profiling is enabled through the flag in Profiler.hpp:

```
./build/./src/main.cpp.o:/Users/ranto/robotics/firmware/src/hello_robot.hpp:42: multiple definition of `prof'; ./build/./src/hello_robot.cpp.o:/Users/ranto/robotics/firmware/src/hello_robot.hpp:42: first defined here
```

This header is included in multiple source files, so we have to move this definition to a source file and access it through the extern in profiler.hpp.

ericv555: I moved the ifdef from the hpp to the cpp, which seems to resolve the error. I think this gets resolved in a more thorough way in my CLI PR where I move the profiler def from profiller.hpp to a makefile macro.

src/sensors/transmitter/dr16.cpp
ericv555: Whoops, this was supposed to be the main thing of the VTM_Comms pull request but I guess I branched off of that one. That PR can be ignored if we do it here.

ericv555: Adjusted pos_x and pos_y to accumulate using += instead of setting them.

src/sensors/transmitter/dr16.cpp
ericv555: Added third variable feed_trigger that expresses the intent to fire that was previously expressed by l_mouse_button. This should stop the feedback loop.

src/sensors/transmitter/dr16.cpp
roccoagain: Ahh gotcha. Then I agree that we should close that PR and just test/merge the VTM changes in this branch.

roccoagain (approved):
LGTM

PR #190, opened 2026-04-22
Command Line Interface

ericv555:
Adds an interactive command line interface that allows for the printing of transmitter data, sensor values, and profiling of each major section of the main loop all as a response to an input command.  Implements a system logger for the purpose of formatting non CLI print statements into the CLI dashboard without interruption. Changes profiler activation from commenting/uncommenting the global prof variable to using make. "make` debug" makes with profiler, "make release" makes without the profiler.


src/main.cpp
```diff
@@ -5,6 +5,7 @@ int main() {
 
     Serial.begin(115200); // the serial monitor is actually always active (for
                           // debug use Serial.println & tycmd)
+	while(!Serial);
```
roccoagain: Could we remove `while (!Serial);`? It blocks startup until a USB serial console is opened. Hive’s USB HID connection doesn’t satisfy this check, so the robot can’t initialize without an open serial monitor.

src/hello_robot.cpp
```diff
@@ -213,7 +233,7 @@ void HelloRobot::check_safety() {
     if (not_safety_mode) {
         // SAFETY OFF
         can.write();
-        // Serial.printf("Can write\n");
+        SystemLog.info(Subsystem::CAN,"Can write\n");
```
roccoagain: It seems that, at 1kHz, "Can write" overwrites the 10 entry history in 10ms, so warnings/errors get drowned out.

src/hello_robot.cpp
```diff
@@ -242,3 +266,302 @@ void HelloRobot::loop_timing() {
...
+				Serial.println("              Examples:");
+				Serial.println("                log motors warn  : Shows motor warnings/errors, and all other system errors");
+				Serial.println("                log all info     : Resets the filter to show absolutely everything");
+                Serial.println();
+                Serial.println("       help");
+                Serial.println("              Displays this manual.");
+}
+void HelloRobot::cmd_live() {
+    num_active_views = 0;
+    SystemLog.is_live_view_active = true;
```
roccoagain, line 446: What happens if you mistype a view name? is_live_view_active is set before the arguments are validated, so `live [typo here]` suppresses logs without opening a dashboard. I suggest setting this after a valid view has been parsed.

src/utils/profiler.hpp
```diff
@@ -14,14 +14,14 @@ struct Profiler {
...
-        /// @brief the time lengths of each profiling section
-        uint32_t time_lengths[PROF_MAX_TIMES] = {0};
         /// @brief Start time for the current profiling section
         uint32_t start_time = 0;
+		/// @brief Average time of section
+		float avg_time = 0.0f;
+		/// @brief max time of section
+        uint32_t max_time = 0;
         /// @brief Number of start/end times recorded
         uint16_t count = 0;
```
roccoagain: At 1 kHz, the 16-bit counter wraps to 0 after about 65.5 seconds, causing the next sample to reset the average and maximum as though it were the first sample (goes down the "first run" branch).

src/hello_robot.cpp
ericv555: Recommented it out

src/utils/profiler.hpp
ericv555: Changed it to a uint32_t should be fine after that

src/hello_robot.cpp
ericv555: made it false before it leaves the function if there are typos

roccoagain (approved):

PR #193, opened 2026-05-25
Ref Update 2026

roccoagain:
Updates to align with RoboMaster’s 2026 spec. Also includes drawing to the player client.

[RMU 2026 Communication Protocol](https://bbs.robomaster.com/wiki/20204847/811363?source=7)

### DIFF
--- a/src/sensors/RefSystem.cpp
+++ b/src/sensors/RefSystem.cpp
@@ -30,7 +30,7 @@ RefSystem::RefSystem() { }
 void RefSystem::init() {
     // clear and start the MCM serial
     MCM_SERIAL.clear();
-    MCM_SERIAL.begin(112500);
+    MCM_SERIAL.begin(115200);
     MCM_SERIAL.flush();
     MCM_SERIAL.clear();
 
--- a/src/sensors/RefSystem.cpp
+++ b/src/sensors/RefSystem.cpp
@@ -248,9 +248,6 @@ void RefSystem::set_ref_data(Frame& frame, uint8_t raw_buffer[REF_MAX_PACKET_SIZ
     case FrameType::EVENT_DATA:
         ref_data.event_data.set_data(frame.data);
         break;
-    case FrameType::PROJECTILE_SUPPLIER_STATUS:
-        ref_data.projectile_supplier_status.set_data(frame.data);
-        break;
     case FrameType::REFEREE_WARNING:
         ref_data.referee_warning.set_data(frame.data);
         break;
@@ -268,9 +265,6 @@ void RefSystem::set_ref_data(Frame& frame, uint8_t raw_buffer[REF_MAX_PACKET_SIZ
         break;
     case FrameType::ROBOT_BUFF:
         ref_data.robot_buff.set_data(frame.data);
-        break;
-    case FrameType::AIR_SUPPORT_STATUS:
-        ref_data.air_support_status.set_data(frame.data);
         break;
     case FrameType::DAMAGE_STATUS:
         ref_data.damage_status.set_data(frame.data);
@@ -310,9 +304,6 @@ void RefSystem::set_ref_data(Frame& frame, uint8_t raw_buffer[REF_MAX_PACKET_SIZ
     case FrameType::SMALL_MAP_COMMAND:
         ref_data.small_map_command.set_data(frame.data);
         break;
-    case FrameType::KBM_INTERACTION:
-        ref_data.kbm_interaction.set_data(frame.data);
-        break;
     case FrameType::SMALL_MAP_RADAR_POSITION:
         ref_data.small_map_radar_position.set_data(frame.data);
         break;
@@ -324,6 +315,16 @@ void RefSystem::set_ref_data(Frame& frame, uint8_t raw_buffer[REF_MAX_PACKET_SIZ
         break;
     case FrameType::SMALL_MAP_ROBOT_DATA:
         ref_data.small_map_robot_data.set_data(frame.data);
+        break;
+    case FrameType::ROBOT_CUSTOM_CONTROLLER_DATA:
+        ref_data.robot_custom_controller_data.set_data(frame.data);
+        break;
+    case FrameType::ROBOT_CUSTOM_CLIENT_DATA:
+        ref_data.robot_custom_client_data.set_data(frame.data);
+        break;
+    case FrameType::CUSTOM_CLIENT_ROBOT_COMMAND:
+        ref_data.custom_client_robot_command.set_data(frame.data);
+        ref_data.keyboard_mouse_control.set_data(frame.data, frame.header.data_length);
         break;
     default:
         Serial.println("Ref System::set_ref_data: Unknown Frame Type");
--- a/src/sensors/RefSystem.cpp
+++ b/src/sensors/RefSystem.cpp
@@ -41,7 +41,7 @@ void RefSystem::read() {
     read_mcm();
 }
 
-void RefSystem::write(uint8_t* packet, uint8_t length) {
+void RefSystem::write(uint8_t* packet, uint16_t length) {
     // return if over baud rate
     if (bytes_sent >= REF_MAX_BAUD_RATE) {
         Serial.println("Too many bytes");
@@ -54,8 +54,33 @@ void RefSystem::write(uint8_t* packet, uint8_t length) {
         return;
     }
 
-    // length of actual sendable data (without the IDs)
-    uint8_t data_length = packet[1] - 6;
+    constexpr uint16_t frame_header_size = FrameHeader::packet_size;
+    constexpr uint16_t command_ID_size = 2;
+    constexpr uint16_t frame_tail_size = 2;
+    constexpr uint16_t interaction_header_size = 6;
+    constexpr uint16_t packet_overhead = frame_header_size + command_ID_size + frame_tail_size;
+
+    if (length < frame_header_size) {
+        Serial.println("Packet Too Short to Send!");
+        return;
+    }
+
+    uint16_t frame_data_length = (static_cast<uint16_t>(packet[2]) << 8) | packet[1];
+    if (frame_data_length > REF_MAX_DATA_SIZE) {
+        Serial.println("Packet Data Too Long to Send!");
+        return;
+    }
+
+    uint16_t expected_length = packet_overhead + frame_data_length;
+    if (length != expected_length) {
+        Serial.println("Packet Length Mismatch!");
+        return;
+    }
+
+    if (frame_data_length < interaction_header_size) {
+        Serial.println("Packet Data Too Short to Send!");
+        return;
+    }
 
     // update header
     packet[0] = 0xA5;                       // set SOF
@@ -67,9 +92,10 @@ void RefSystem::write(uint8_t* packet, uint8_t length) {
     packet[10] = ref_data.robot_performance.robot_ID >> 8;
 
     // update tail
-    uint16_t footerCRC = generateCRC16(packet, 13 + data_length);
-    packet[13 + data_length] = (footerCRC & 0x00FF);    // set CRC
-    packet[14 + data_length] = (footerCRC >> 8);        // set CRC
+    uint16_t footer_index = packet_overhead + frame_data_length - frame_tail_size;
+    uint16_t footerCRC = generateCRC16(packet, footer_index);
+    packet[footer_index] = (footerCRC & 0x00FF);    // set CRC
+    packet[footer_index + 1] = (footerCRC >> 8);    // set CRC
 
     if (Serial7.write(packet, length) == length) {
         packets_sent++;
--- a/src/sensors/RefSystem.cpp
+++ b/src/sensors/RefSystem.cpp
@@ -36,7 +36,7 @@ void RefSystem::init() {
 
     // clear and start the VTM serial
     VTM_SERIAL.clear();
-    VTM_SERIAL.begin(112500);
+    VTM_SERIAL.begin(921600);
     VTM_SERIAL.flush();
     VTM_SERIAL.clear();
 }
--- a/src/sensors/RefSystem.cpp
+++ b/src/sensors/RefSystem.cpp
@@ -2,6 +2,11 @@
 #include "RefSystemPacketDefs.hpp"
 #include "comms/data/sendable.hpp"
 
+// Uncomment to enable debug prints
+#define REF_SYSTEM_DEBUG
+// Uncomment to debug even harder
+// #define REF_SYSTEM_DEBUG_ALL_FRAMES
+
 uint8_t generateCRC8(uint8_t* data, uint32_t len) {
     uint8_t CRC8 = 0xFF;
     while (len-- > 0) {
@@ -153,8 +158,9 @@ bool RefSystem::read_frame_header(HardwareSerial* serial, uint8_t raw_buffer[REF
     frame.header.CRC = raw_buffer[buffer_index + 4];
 
     // verify the CRC is correct
-    if (frame.header.CRC != generateCRC8(raw_buffer, 4)) {
-        Serial.println("Header failed CRC");
+    uint8_t expected_CRC = generateCRC8(raw_buffer, 4);
+    if (frame.header.CRC != expected_CRC) {
+        Serial.printf("[Ref] Header CRC failed: received=0x%02x expected=0x%02x\n", frame.header.CRC, expected_CRC);
         packets_failed++;
         return false;
     }
@@ -183,7 +189,7 @@ bool RefSystem::read_frame_command_ID(HardwareSerial* serial, uint8_t raw_buffer
 
     // sanity check, verify the ID is valid
     if (frame.commandID > REF_MAX_COMMAND_ID) {
-        Serial.println("Invalid Command ID");
+        Serial.printf("[Ref] Invalid command ID: 0x%04x\n", frame.commandID);
         packets_failed++;
         return false;
     }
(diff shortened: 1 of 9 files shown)


src/sensors/RefSystemPacketDefs.hpp
```diff
@@ -70,6 +67,12 @@ enum FrameType {
     SMALL_MAP_SENTRY_COMMAND = 0x0307,
     /// @brief Robot data received by player clients' Small Map
     SMALL_MAP_ROBOT_DATA = 0x0308,
+    /// @brief Custom data sent from a robot to the Custom Controller.
+    ROBOT_CUSTOM_CONTROLLER_DATA = 0x0309,
+    /// @brief Custom data sent from a robot to the Custom Client.
+    ROBOT_CUSTOM_CLIENT_DATA = 0x0310,
+    /// @brief Custom command sent from the Custom Client to robots.
+    CUSTOM_CLIENT_ROBOT_COMMAND = 0x0311,
 };
```
ericv555, line 94: Do we need the radar frames? I.E frames 0x0A01 through 0x0A06.

src/sensors/RefSystemPacketDefs.hpp
```diff
@@ -173,6 +174,7 @@ struct GameStatus {
         competition_type = data[0] & 0x0F;
         current_stage = (data[0] >> 4) & 0x0F;
         round_time_remaining = (data[2] << 8) | data[1];
+        unix_time = 0;
```
ericv555, line 202: unix_time already set to zero in the initial variable definition. You can remove this line

src/sensors/RefSystemPacketDefs.hpp
```diff
@@ -1,12 +1,15 @@
...
+/// @brief Maximum size of a Ref System packet in bytes \n
+/// @brief 5 byte header + 2 byte command ID + 300 + 2 byte CRC
+constexpr uint16_t REF_MAX_PACKET_SIZE = 309;
+// TODO: This size is only required by 0x0310 and pretty excessive to use for every packet. Maybe we could reduce the raw array length for most packets or rework how the raw bytes are stores in each struct.
 
-/// @brief Maximum size of a Ref System packet in bytes
-constexpr uint16_t REF_MAX_PACKET_SIZE = 128;
 /// @brief Maximum valid command ID for Ref System packets
-constexpr uint16_t REF_MAX_COMMAND_ID = 0x0308;
+constexpr uint16_t REF_MAX_COMMAND_ID = 0x0311;
```
ericv555: The largest packet is 0x0310 not 0x311

src/sensors/RefSystem.cpp
ericv555:
```suggestion
void RefSystem::write(uint8_t* packet, uint16_t length) {
```
length can now be up to 309 which exceeds uint8_t max size.

src/sensors/RefSystemPacketDefs.hpp
```diff
@@ -1620,6 +1613,320 @@ struct SmallMapRobotData {
...
+        }
+    }
+};
+
+/// @brief Custom command sent from the Custom Client to robots.
+/// @note transmitted at a maximum frequency of 75 Hz via VTM link
+/// @note ID: 0x0311
+struct CustomClientRobotCommand {
+    /// @brief Size of the CustomClientRobotCommand packet in bytes
+    static const uint16_t packet_size = 30;
```
ericv555:
```suggestion
    static const uint8_t packet_size = 30;
```
Those 8 bits could come in handy later...

src/sensors/RefSystemPacketDefs.hpp
```diff
@@ -244,60 +246,65 @@ struct GameResult {
...
-        data.blue_engineer_health = blue_team_HP[1];
-        data.blue_standard_3_health = blue_team_HP[2];
-        data.blue_standard_4_health = blue_team_HP[3];
-        data.blue_standard_5_health = blue_team_HP[4];
-        data.blue_sentry_health = blue_team_HP[5];
+        data.hero_health = hero_health;
+        data.engineer_health = engineer_health;
+        data.standard_3_health = standard_3_health;
+        data.standard_4_health = standard_4_health;
+        data.sentry_health = sentry_health;
```
ericv555, line 332: Is there a reason why we wouldn't send outpost and base health to comms?

src/sensors/RefSystemPacketDefs.hpp
```diff
@@ -328,8 +335,8 @@ struct EventData {
     /// @param data FrameData object to extract data from
     void set_data(FrameData data) {
         memcpy(raw, data.data, packet_size);
-        reload_zone_status = (data[0] >> 2) & 0x01;   // bit 2
-        capture_point_status = (data[2] >> 5) & 0x03; // bits 21 and 22
+        reload_zone_status = (data[0] >> 2) & 0x01;                               // bit 2
+        capture_point_status = ((data[2] >> 7) & 0x01) | ((data[3] & 0x01) << 1); // bits 23 and 24
     }
```
ericv555, line 365: There is a ton more data we could collect from this packet if software would find it useful. You could add a TODO or put up a github issue for it.

src/sensors/RefSystemPacketDefs.hpp
ericv555: You didnt add this but you could remove it if you want. Seems like it is not marked as reserved on the doc.

src/sensors/RefSystemPacketDefs.hpp
```diff
@@ -552,69 +520,61 @@ struct RobotPerformance {
...
     /// @brief Fills in this struct with the data from a FrameData object
     /// @param data FrameData object to extract data from
     void set_data(FrameData data) {
         memcpy(raw, data.data, packet_size);
-        chassis_voltage_output = (data[1] << 8) | data[0];
-        chassis_current_output = (data[3] << 8) | data[2];
+        reserved_1 = (data[1] << 8) | data[0];
+        reserved_2 = (data[3] << 8) | data[2];
+
+        uint32_t reserved_3_raw = (data[7] << 24) | (data[6] << 16) | (data[5] << 8) | data[4];
```
ericv555: idk if this will work or not integer promotion rules are confusing ( im pretty sure that by shifting data which is a uint8_t more then 8 spaces in this line and by implicitly casting it to a larger size, it will be promoted to a signed int but im really not sure). You can guarantee that there will be not be problems with this line by explicitly casting each uint8_t data[i] to a uint32 before shifting.

src/sensors/RefSystemPacketDefs.hpp
ericv555, line 835: Should we share num_17mm_fortress with comms?

src/sensors/RefSystemPacketDefs.hpp
```diff
@@ -895,76 +836,131 @@ struct RFIDStatus {
...
+        our_tunnel_trapezoid_lower = (rfid_status >> 29) & 0x01;
+        our_tunnel_trapezoid_middle = (rfid_status >> 30) & 0x01;
+        our_tunnel_trapezoid_upper = (rfid_status >> 31) & 0x01;
+        their_tunnel_road_lower = data[4] & 0x01;
+        their_tunnel_road_middle = (data[4] >> 1) & 0x01;
+        their_tunnel_road_upper = (data[4] >> 2) & 0x01;
+        their_tunnel_trapezoid_lower = (data[4] >> 3) & 0x01;
+        their_tunnel_trapezoid_middle = (data[4] >> 4) & 0x01;
+        their_tunnel_trapezoid_upper = (data[4] >> 5) & 0x01;
+        reserved = (data[4] >> 6) & 0x03;
```
ericv555, line 978: You may want to comment and say that this isn't explicitly reserved. The 'reserved data' is just unused.

src/sensors/RefSystemPacketDefs.hpp
ericv555: I confirmed with some test code that this may have an unexpected outcome. each instance of data should be static_cast<uint32_t>(data[i])

src/sensors/RefSystemPacketDefs.hpp
```diff
@@ -1,12 +1,18 @@
 #pragma once
 
-#include <Arduino.h>
 #include "comms/data/comms_ref_data.hpp"
+#include <Arduino.h>
+
+/// @brief Maximum size of the data segment in a Ref System packet. The record is currently held by 0x0301
```
ericv555: The largest packet is 0x0310 not 0x301

src/sensors/RefSystemPacketDefs.hpp
```diff
@@ -244,60 +253,65 @@ struct GameResult {
...
 
     /// @brief Fills in this struct with the data from a FrameData object
     /// @param data FrameData object to extract data from
     void set_data(FrameData data) {
         memcpy(raw, data.data, packet_size);
-        for (int i = 0; i < 8; i++) {
-            red_team_HP[i] = (data[2 * i + 1] << 8) | data[2 * i];
-            blue_team_HP[i] = (data[16 + 2 * i + 1] << 8) | data[16 + 2 * i];
-        }
+        hero_health = (data[1] << 8) | data[0];
```
ericv555, line 314: static cast data to uint16_t and complete for all further lines of set_data

src/sensors/RefSystemPacketDefs.hpp
```diff
@@ -674,67 +641,37 @@ struct RobotBuff {
...
-        Serial.printf("\tTime Remaining: %u\n", time_remaining);
-    }
-
-    /// @brief Fills in this struct with the data from a FrameData object
-    /// @param data FrameData object to extract data from
-    void set_data(FrameData data) {
-        memcpy(raw, data.data, packet_size);
-        status = data[0];
-        time_remaining = data[1];
+        heat_cooling = (data[2] << 8) | data[1];
```
ericv555, line 685: static cast data to a uint16_t

src/sensors/RefSystemPacketDefs.hpp
```diff
@@ -1077,58 +1073,86 @@ struct GroundRobotPositions {
         memcpy(&standard_4_x, &standard_4_x_raw, sizeof(standard_4_x));
         uint32_t standard_4_y_raw = (data[31] << 24) | (data[30] << 16) | (data[29] << 8) | data[28];
```
ericv555: static_cast data to a uint32_t

src/sensors/RefSystemPacketDefs.hpp
```diff
@@ -1077,58 +1073,86 @@ struct GroundRobotPositions {
...
     /// @param data FrameData object to extract data from
     void set_data(FrameData data) {
         memcpy(raw, data.data, packet_size);
-        hero = data[0];
-        engineer = data[1];
-        standard_3 = data[2];
-        standard_4 = data[3];
-        standard_5 = data[4];
-        sentry = data[5];
+        uint16_t tracking_progress = (data[1] << 8) | data[0];
```
ericv555, line 1157: while not strictly neccesary here since shifting 8 in a uint16 wont effect the signed bit. I would probably still static cast this to a uint16

src/sensors/RefSystemPacketDefs.hpp
```diff
@@ -1137,26 +1161,30 @@ struct RadarProgress {
...
         Serial.println("SentryDecision:");
         Serial.printf("\tSentry Info: %u\n", sentry_info);
+        Serial.printf("\tSentry Info 2: %u\n", sentry_info_2);
     }
 
     /// @brief Fills in this struct with the data from a FrameData object
     /// @param data FrameData object to extract data from
     void set_data(FrameData data) {
         memcpy(raw, data.data, packet_size);
         sentry_info = (data[3] << 24) | (data[2] << 16) | (data[1] << 8) | data[0];
```
ericv555: Static cast is necessary here

src/sensors/RefSystemPacketDefs.hpp
```diff
@@ -1271,7 +1299,7 @@ struct CustomControllerRobot {
 /// @note ID: 0x0303
 struct SmallMapCommand {
     /// @brief Size of the SmallMapCommand packet in bytes
-    static const uint8_t packet_size = 15;
+    static const uint8_t packet_size = 12;
```
ericv555, line 1335: There is a mismatch in the Robomaster doc on this. There is only 12 bytes of data that they say they will send, but in the list at the top of the doc with the sizes it says that this one is 15 bytes.

src/sensors/RefSystemPacketDefs.hpp
```diff
@@ -1620,6 +1620,320 @@ struct SmallMapRobotData {
...
+            Serial.printf("\tData[%u]: %u\n", i, data[i]);
+        }
+    }
+
+    /// @brief Fills in this struct with the data from a FrameData object
+    /// @param data FrameData object to extract data from
+    void set_data(FrameData data) {
+        memcpy(raw, data.data, packet_size);
+        for (uint16_t i = 0; i < packet_size; i++) {
+            this->data[i] = data[i];
```
ericv555:
```suggestion
            this->data[i] = data.data[i];
```
data is a FrameData object data.data is a uint8_t array

src/sensors/RefSystemPacketDefs.hpp
```diff
@@ -1620,6 +1620,320 @@ struct SmallMapRobotData {
...
```
ericv555:
```suggestion
            this->data[i] = data.data[i];
```

src/sensors/RefSystemPacketDefs.hpp
```diff
@@ -1620,6 +1620,320 @@ struct SmallMapRobotData {
...
```
ericv555:
```suggestion
            this->data[i] = data.data[i];
```

src/sensors/RefSystemPacketDefs.hpp
```diff
@@ -1620,6 +1620,320 @@ struct SmallMapRobotData {
...
+    static const uint32_t timeout_ms = 250;
+
+    /// @brief x-axis moving speed of the mouse. A negative value indicates a left movement.
+    int32_t mouse_speed_x = 0;
+    /// @brief y-axis moving speed of the mouse. A negative value indicates a downward movement.
+    int32_t mouse_speed_y = 0;
+    /// @brief Scroll wheel's moving speed of the mouse. A negative value indicates a backward movement.
+    int32_t scroll_speed = 0;
+    /// @brief Whether the mouse's left button is pressed.
+    uint8_t button_left = 0;
```
ericv555: button_left right and middle should be bools instead

ericv555 (changes requested):
Looks solid 1 or 2 big bugs most other comments are small optimizations.

src/sensors/RefSystemPacketDefs.hpp
roccoagain: I'm leaving those those out because there is no radar in RMUL.

src/sensors/RefSystem.cpp
roccoagain: Holy good catch

src/sensors/RefSystemPacketDefs.hpp
roccoagain: Those are not present in RMUL. I guess the same could be said about the engineer, but that was already in `to_comms_data` so I left it.

src/sensors/RefSystemPacketDefs.hpp
roccoagain: lol true

src/sensors/RefSystemPacketDefs.hpp
roccoagain: Yep, this was strange. The members of `map_command_t` sum to 12 bytes, but the packet is listed at 15 bytes in that first table. I'm going with 12 bytes because it's unlikely that they left out 3 bytes. I will make note of this though.

src/sensors/RefSystemPacketDefs.hpp
roccoagain: I did this so that the bits from the last packet would not carry over. `unix_time` is built with `|=` and `<<=` so I think it's necessary to start from 0 each time.

src/sensors/RefSystemPacketDefs.hpp
roccoagain: Nope, there is no fortress in RMUL

src/sensors/RefSystemPacketDefs.hpp
roccoagain: I believe this is equivalent, since the `[]` operator is overloaded in `struct FrameData`.

Defined as:
`uint8_t operator[](int index) { return data[index]; }`
`uint8_t operator[](int index) const { return data[index]; }`

src/sensors/RefSystemPacketDefs.hpp
roccoagain: I believe this is equivalent, since the `[]` operator is overloaded in `struct FrameData`.

Defined as:
`uint8_t operator[](int index) { return data[index]; }`
`uint8_t operator[](int index) const { return data[index]; }`

src/sensors/RefSystemPacketDefs.hpp
roccoagain: I believe this is equivalent, since the `[]` operator is overloaded in `struct FrameData`.

Defined as:
`uint8_t operator[](int index) { return data[index]; }`
`uint8_t operator[](int index) const { return data[index]; }`

src/sensors/RefSystemPacketDefs.hpp
roccoagain: It looks like the other fields are exclusive to RMUC gameplay. Did you notice anything in particular?

ericv555: Nope I was looking at the RMUC stuff my bad.

roccoagain:
Tested last week; keyboard/mouse driving and ref drawing are both working. Idling looks healthy so far, but we still need to test with a full field/game setup.

Sammyboyo (approved):
the clock strikes

Sammyboyo (approved):

PR #194, opened 2026-06-01
Camera state synchronization and other stuff

Sammyboyo:


src/sensors/StereoCamTrigger.cpp
```diff
@@ -35,14 +45,34 @@ void StereoCamTrigger::init() {
...
+    digitalWrite(config.camera_2_line_2_pin, LOW);
+
+    Serial.printf("counter reset pin: %u triggered\n", config.camera_1_line_2_pin);
+  }
+  comms_layer.get_hive_data().stereo_cam_start_stop.stop_received = false;
+  comms_layer.get_hive_data().stereo_cam_start_stop.start_received = false;
+}
+
 void StereoCamTrigger::send_to_comms() const {
   Comms::Sendable<StereoCamTriggerData> sendable;
```
roccoagain: `track_exposures()` writes to `comms_data.state` in the timer interrupt, but `send_to_comms()` copies `comms_data` from the main loop. Is that a freaking data race, yo? I think we should take a snapshot of comms_data with interrupts disabled, then send that.

roccoagain (approved):
Image

PR #195, opened 2026-06-05
Non Blocking Sensors

ericv555:
The purpose of this PR is to convert the ET16S, ICM IMU, and buffencoders from blocking sensor reads to non-blocking sensor reads. Previously, when the read function for any of these objects was called, the teensy would halt (block) and wait to receive the data. After it had waited for all the data to come in, it would do whatever it needed to with it. This PR changes that, so when one of these objects' read statement is called, it already has the data in its buffer, and it can immediately process it. This is achieved using DMA transfers that are built into the Teensy but have been unused until now. This had to be done in 3 different ways for the 3 different objects that I worked on, so I will go through them one at a time.
* ET16S DMA Implementation
The eDMA channel is set up in the init, where an interrupt is attached that triggers whenever a full ET16S frame has been sent over the wire. This interrupt calls the ET16S interrupt service routine (ISR) function, which sets the buffer that has just been filled to the active buffer that can be read in the read function. The ISR then sets a second buffer as the new target buffer, which immediately begins filling via the DMA channel.  These two ping pong buffers trade back and forth whenever one of them fills.
* ICM IMU DMA Implementation
Requires a request_read function in sensor_manager that is called in the main loop. request_read starts an SPI transfer and continues on while the transfer takes place. An EventResponder object is attached to the SPI transfer that returns true when it is complete. This is checked in the regularly placed read function, where it then sends data to comms if the data transfer has completed. This way, the ICM IMU requires no interrupt.
* Buff Encoder DMA Implementation
This one was the hardest to implement since our robots have 3 or 4 encoders on them. When the sensor_manager calls request_read, it kicks off a DMA transfer on a separate SPI bus, then the ICM IMU (the IMU uses SPI1 and the buff encoder uses SPI, so the IMU and the buff encoders can read data at the same time). When the first buff_encoder transfer finishes, it triggers an ISR function that interrupts the main loop, ends the SPI transaction and starts the SPI transaction with the next encoder. It does this until all encoders have completed their transaction, where it then triggers a bool flag for transfer completion.


src/sensors/sensor_manager.cpp
```diff
@@ -1,37 +1,64 @@
...
+        pinMode(enc.spi_cs, OUTPUT);
+        digitalWrite(enc.spi_cs, HIGH);
+    }
+    for (const auto& imu : config_data.icm_imus) {
+        pinMode(imu.spi_cs, OUTPUT);
+        digitalWrite(imu.spi_cs, HIGH);
+    }
+    SPI1.setMISO(config_data.icm_imus[0].spi_miso);
+	SPI1.setMOSI(config_data.icm_imus[0].spi_mosi);
+	SPI1.setSCK(config_data.icm_imus[0].spi_sck);
```
roccoagain: On a robot config with no ICM IMUs, this is undefined behavior (trying to read element 0, but the array is empty). This may be the case in the future, if we add a new IMU and remove the `Icm Imus` block from the YAML.

Maybe:
```
if (!config_data.icm_imus.empty()) {
    const auto& imu = config_data.icm_imus[0];
    SPI1.setMISO(imu.spi_miso);
    SPI1.setMOSI(imu.spi_mosi);
    SPI1.setSCK(imu.spi_sck);
} else {
    SystemLog.error(Subsystem::SENSORS, "No ICM IMUs configured; SPI1 pins not set\n");
}
SPI1.begin();
```

src/sensors/sensor_manager.cpp
```diff
@@ -40,7 +67,9 @@ void SensorManager::configure_sensors(const Cfg::RobotConfig& config_data) {
 
     for (const auto& icm_config : config_data.icm_imus) {
         Serial.printf("Configuring ICM20649 with name %u\n", static_cast<uint32_t>(icm_config.imu_name));
-        sensors.emplace(icm_config.imu_name, std::make_shared<ICM20649>(icm_config));
+		auto imu_ptr = std::make_shared<ICM20649>(icm_config);
+        sensors.emplace(icm_config.imu_name, imu_ptr);
+		icm_imu = imu_ptr;
```
roccoagain, line 80: `icm_imu` is reassigned on every pass through this loop, sow hen it ends it only points to the last ICM.

`request_read()` only starts a SPI transfer for icm_imu, so only the last one will get new data. This is coincidentally fine for today's config, because we only have one ICM.

roccoagain (commented):
Partial review; leaving these comments now while I continue to review.

src/sensors/sensor_manager.cpp
ericv555: Nice catch.

src/sensors/sensor_manager.cpp
ericv555: True. One of the downsides of using non blocking code is it makes changes like adding a second icm_imu a bit more difficult. I designed the interrupt logic for the imu assuming there will only be 1; if we have the option to have multiple, an entirely different logic will need to be used. Probably using the eDMA channels' ability to chain interrupts, like what I did for the buff encoders. For the time being, I will put a comment here saying that this will only work with 1 icm imu I'll also make it safety_procedure.

src/sensors/buff_encoder.hpp
```diff
@@ -99,4 +115,21 @@ class BuffEncoder : public Sensor {
...
     static const SPISettings m_settings;
+	
+	/// @brief Buffer of transmitted data to the buff encoders
+    alignas(32) uint8_t tx_buffer[32];
+	
+	/// @brief Buffer of recieved data from the buff encoders
+    alignas(32) uint8_t rx_buffer[32];
+
+	/// @brief Pointer to the SensorManager's active transfer flag
+    const volatile bool* shared_dma_flag;
```
roccoagain: `shared_dma_flag` and `rx_buffer` are uninitialized, so the `!= nullptr` guard in `read()` doesn't work properly. Default them to `nullptr` and `{0}`.

src/sensors/buff_encoder.cpp
```diff
@@ -10,38 +12,53 @@ void BuffEncoder::init() {
...
         write_zero_pos(0);
         read_count++;
     }
     read_count = 0;
 }
+void BuffEncoder::isr_start_transfer(EventResponderRef spi_event) {
+	SPI.beginTransaction(m_settings);
+	digitalWrite(config_data.spi_cs, LOW);
 
+	SPI.transfer(tx_buffer, rx_buffer, 6, spi_event); //after testing make this an assert_or_safety_procedure()
```
roccoagain: SPI.transfer returns a bool, which you could use to log or safety assert here

You already left a comment, just leaving this here as a reminder lol

src/sensors/transmitter/ET16S.cpp
```diff
@@ -27,56 +40,120 @@ void ET16S::init() {
 	channel[3].kind = InputKind::STICK;
 	//configure remaining channels
 	set_config();
+	setup_edma_channel();
 }
+void ET16S::setup_edma_channel() {
+    // Disable the HardwareSerial CPU RX interrupt so that it doesnt compete with dma interrupt
+    LPUART5_CTRL &= ~LPUART_CTRL_RIE;
```
roccoagain: Serial8.begin() sets ILIE too (CTRL_ENABLE in HardwareSerial.cpp:80), perhaps clear ILIE here as well.

roccoagain (approved):

roccoagain (approved):

PR #196, opened 2026-06-07
State desync fr this time

Sammyboyo:
adds a counter that tracks the number of frames that have been triggered which is what we use to sync the robot state with a given camera frame.

Most of the code is actually a new comms packet that is used to test round trip latency of comms.

### DIFF
--- a/src/comms/data/test_data.hpp
+++ b/src/comms/data/test_data.hpp
@@ -29,6 +29,7 @@ struct BigTestData : Comms::CommsData {
     float blah[128] = { 0 };
 };
 
+/// @brief Struct used to measure the round trip latency of the comms system
 struct TestLatencyData : Comms::CommsData {
     TestLatencyData() : Comms::CommsData(Comms::TypeLabel::TestLatencyData, Comms::PhysicalMedium::Ethernet, Comms::Priority::High, sizeof(TestData)) {}
     /// @brief The time in micros when the packet was sent
--- a/src/comms/data/test_data.hpp
+++ b/src/comms/data/test_data.hpp
@@ -31,7 +31,7 @@ struct BigTestData : Comms::CommsData {
 
 /// @brief Struct used to measure the round trip latency of the comms system
 struct TestLatencyData : Comms::CommsData {
-    TestLatencyData() : Comms::CommsData(Comms::TypeLabel::TestLatencyData, Comms::PhysicalMedium::Ethernet, Comms::Priority::High, sizeof(TestData)) {}
+    TestLatencyData() : Comms::CommsData(Comms::TypeLabel::TestLatencyData, Comms::PhysicalMedium::Ethernet, Comms::Priority::High, sizeof(TestLatencyData)) {}
     /// @brief The time in micros when the packet was sent
     uint64_t current_time = 0;
     /// @brief the difference in time between the current time and the time that was contained in the last received packet
--- a/src/comms/data/test_data.hpp
+++ b/src/comms/data/test_data.hpp
@@ -28,3 +28,11 @@ struct BigTestData : Comms::CommsData {
     /// @brief a bunch of data
     float blah[128] = { 0 };
 };
+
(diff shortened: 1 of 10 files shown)


src/sensors/StereoCamTrigger.cpp
```diff
@@ -17,9 +17,12 @@ void StereoCamTrigger::track_exposures() {
   digitalWrite(config.digital_trigger_pin_1, LOW);
   digitalWrite(config.digital_trigger_pin_2, LOW);
   
+  counter += 1;
+
   if(estimated_state_map_interrupt_safe != nullptr) {
     // copy the estimated state map to the local estimated state map
     estimated_state_map_interrupt_safe->fill_state_array(comms_data.state);
+    comms_data.frame_count = counter;
```
roccoagain, line 25: This callback updates `comms_data` while the main loop could be copying it in `send_to_comms()`.

src/comms/data/test_data.hpp
```diff
@@ -27,4 +27,13 @@ struct BigTestData : Comms::CommsData {
 
     /// @brief a bunch of data
     float blah[128] = { 0 };
-};
\ No newline at end of file
+};
+
+/// @brief Struct used to measure the round trip latency of the comms system
+struct TestLatencyData : Comms::CommsData {
+    TestLatencyData() : Comms::CommsData(Comms::TypeLabel::TestLatencyData, Comms::PhysicalMedium::Ethernet, Comms::Priority::High, sizeof(TestData)) {}
```
roccoagain: I think `sizeof(TestData)` should be `sizeof(TestLatencyData)`

roccoagain (dismissed):

src/sensors/StereoCamTrigger.hpp
```diff
@@ -34,6 +34,9 @@ class StereoCamTrigger : public Sensor{
 
     /// @brief timestamp of the last time an exposure was triggered (last time signal was set to HIGH)
     volatile uint32_t latest_exposure_timestamp = 0;
+
+    /// @brief the number of frames that have been triggered since the last counter reset.
+    int counter = -1;
```
ericv555:
```suggestion
   volatile int counter = -1;
```
Variables modified in interrupts and accessed outside of them should be marked as volatile so the compiler doesn't optimize them out.

src/sensors/StereoCamTrigger.cpp
```diff
@@ -78,4 +83,4 @@ void StereoCamTrigger::send_to_comms() const {
 
   sendable.data = comms_data;
```
ericv555, line 85:
```suggestion
  noInterrupts();
  sendable.data = comms_data;
  interrupts();
```
I would prevent interrupts from going off while comms_data is being copied. Otherwise, comms_data could be interrupted and modified in the middle of this copy.

src/sensors/StereoCamTrigger.cpp
Sammyboyo: I believe that it is impossible for the send to comms to ever get a garbage value because setting an int is only 1 instruction, and either getting the old or new value is chill I think.

src/sensors/StereoCamTrigger.cpp
Sammyboyo: I believe that Rishi and I tried that but we got some weird slow loop times because of it and it seems to work ok without it, but we should probably look into it some more.

ericv555: I would try it again and make sure it is only that one line in it. The assignment to sendable shouldn't be more than 7 or 8 instructions, so it shouldn't cause a slowdown. If the send to comms line was in the noInterrupts it may cause a problem. If it doesnt work its no big deal though the assignment is so fast its probably really unlikely for the interrupt to go off during it.

Sammyboyo: has been added

roccoagain (approved):
APPROVED!

PR #197, opened 2026-06-09
Accept override state for gimbal yaw

effdotsh:
Needed for lidar localization


roccoagain (approved):
Screenshot 2026-06-08 at 8 51 28 PM

PR #199, opened 2026-06-22
Delay until valid packet for yaw

effdotsh:


src/sensors/buff_encoder.cpp
```diff
@@ -3,13 +3,26 @@
...
+    }
+    read_count = 0;
 }
 
 void BuffEncoder::read() {
+    while (read_zero_pos() != 0.0f && read_count < read_zero_pos_max_attempts) {
+        write_zero_pos(0);
+        read_count++;
+    }
+    read_count = 0;
```
roccoagain, line 25: BuffEncoder::read() runs in the 1 kHz main loop, but this retry counter sets `read_count = 0` every call, so a persistent nonzero zero position can trigger 10 attempts every millisecond.

src/sensors/buff_encoder.cpp
```diff
@@ -25,19 +38,141 @@ void BuffEncoder::read() {
     digitalWrite(config_data.spi_cs, HIGH);
     SPI.endTransaction();
 
+    uint8_t status = data[4] & 0x07;
+    uint8_t crc_received = data[5];
+    uint8_t crc_computed  = mt6835_crc8(&data[2], 3);
+
+    if (crc_received != crc_computed) {
+        Serial.printf("Pin: %u, MT6835 CRC mismatch\n", config_data.spi_cs);
```
roccoagain, line 46: On CRC mismatch, this logs the error but still decodes and publishes the failed frame. Return early or keep the previous valid angle.

src/sensors/buff_encoder.cpp
Sammyboyo: That should not happen and if it does something is very wrong. In practice the write only gets called once on some boots if the zero pos register got flipped, but we check in the read loop just in case. I could be convinced to lower the counter if you think it is too slow.

src/sensors/buff_encoder.cpp
Sammyboyo: true

Sammyboyo (approved):

src/sensors/buff_encoder.cpp
roccoagain: Seems legit

PR #200, opened 2026-06-23
adds 3 flywheel option to the flywheel controller

Sammyboyo:


src/controls/controller.cpp
```diff
@@ -390,24 +390,30 @@ void FlywheelController::step(RobotStateMap& reference_map, RobotStateMap& estim
 
     float target_motor_velocity = pid_high.filter(dt, false, false) * controller_config.gear_ratios.ball_to_flywheel_rad;
     
-    std::shared_ptr<Motor> flywheel_motors[2] = { flywheel_motor_1, flywheel_motor_2 };
-    float motor_outputs[2];
+    std::shared_ptr<Motor> flywheel_motors[3] = { flywheel_motor_1, flywheel_motor_2, flywheel_motor_3 };
```
ericv555, line 393: I'd make this static or move flywheel_motors to be a class member. no need to instantiate that every step. Probably should be a class member and set motor_1 2 and 3 in the constructor or in the member definition

src/controls/controller.cpp
```diff
@@ -390,24 +390,30 @@ void FlywheelController::step(RobotStateMap& reference_map, RobotStateMap& estim
...
+    float motor_outputs[3];
 
-    for (int i = 0; i < 2; i++) {
+    for (int i = 0; i < num_flywheel_motors; i++) {
```
ericv555, line 395:
```suggestion
    for (int i = 0; i < num_flywheel_motors; i++) {
     if (!flywheel_motors[i]) {
            pid_low.measurement = 0.0f;
			continue; 
	 }
```
You can replace the 2 if statements in this loop with 1 by adding this. The shared pointer should convert to bool in an if statement.

src/controls/controller.cpp
```diff
@@ -390,24 +390,30 @@ void FlywheelController::step(RobotStateMap& reference_map, RobotStateMap& estim
...
+        } else {
+            pid_low.measurement = 0.0f;
+        }
         motor_outputs[i] = pid_low.filter(dt, true, false);
     }
 
     flywheel_motor_1->write_motor_torque(motor_outputs[0]);
     flywheel_motor_2->write_motor_torque(motor_outputs[1]);
+    if (flywheel_motor_3 != nullptr) {
+        flywheel_motor_3->write_motor_torque(motor_outputs[2]);
```
ericv555: You can get rid of these 3 writes and do them all in one line since you have that flywheel_motor array. you would add this to the loop
```
	flywheel_motors[i]->write_motor_torque(motor_outputs[i]);
```

src/controls/controller.cpp
```diff
@@ -390,24 +390,30 @@ void FlywheelController::step(RobotStateMap& reference_map, RobotStateMap& estim
...
         pid_low.kf = low_level_velocity_controller.gains.f;
 
-        pid_low.setpoint = i == 0 ? target_motor_velocity * controller_config.gear_ratios.motor1_direction : 
-                                    target_motor_velocity * controller_config.gear_ratios.motor2_direction;
+        pid_low.setpoint = target_motor_velocity * flywheel_directions[i];
 
-        pid_low.measurement = flywheel_motors[i]->get_state().speed;
+        if (flywheel_motors[i] != nullptr) {
+            pid_low.measurement = flywheel_motors[i]->get_state().speed;
+        } else {
```
ericv555: You can get rid of this if-else statement with the other comments I suggested.

ericv555 (commented):
Looks good

Pandabear1125 (approved):
LGTM

PR #201, opened 2026-06-26
Slow update yaw error

effdotsh:
let lidar correct imu slowly to allow autoaim to still have useful relative yaw values


roccoagain (approved):
Looks good perchance

PR #209, opened 2026-08-19
Improvements to Makefile

roccoagain:
Rework of the Makefile:
- Parallel builds actually work now. `-j$(nproc)` was a no-op on macOS since `nproc` doesn't exist there. `make JOBS=1` if you want a serial output while building.
- `git_info.h` is only rewritten when its contents change. It was being rebuilt every `make`.
- `git_scraper` is compiled once into `build/` instead of being rebuilt and deleted every build.
- Per-object disassembly is opt-in behind `DUMP_OBJS=1`. It was adding ~600MB to `build/` and a few seconds to every clean build.
- Raplaced `-g3` with `-g2`. `-g3` emits macro definitions, which cost object size and aren't used by the addr2line in `teensy4/CrashReport.cpp`.
- files in `src/` had includes that only resolved because every subdirectory of `src/` was on the search path. They're relative to `src/` now.
- `LIBRARY_INC_DIRS` no longer uses `find -maxdepth 2`, which pulled in directories that aren't include roots and emitted malformed `-isystem` paths.
- Separate `.elf`, `.hex`, and `dump` targets
- `FEATURE_DEFINES` replaces the commented-out `#define`s scattered in source files. Set it at the top of the Makefile.
- Build artifacts stay in build/ now, no more `firmware.hex` and `firmware.elf` in the repo root
- clangd is now decoupled from normal builds. Run `make clangd` if you use clangd.

### DIFF
--- a/tools/generate_compile_commands.sh
+++ b/tools/generate_compile_commands.sh
@@ -39,8 +39,6 @@ get_system_includes() {
 : "${CFLAGS:?CFLAGS is required}"
 : "${SRC_FILES:?SRC_FILES is required}"
 
-echo "[Generating compile_commands.json]"
-
 if [[ ! -x "$COMPILER_CPP" ]]; then
     echo "Error: C++ compiler not found at $COMPILER_CPP" >&2
     exit 1
@@ -138,5 +136,3 @@ printf '\n]\n' >&3
 exec 3>&-
 mv "$tmp_db" compile_commands.json
 trap - EXIT
-
-echo "[compile_commands.json generated with $entry_count entries]"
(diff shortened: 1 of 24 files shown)


tools/generate_compile_commands.sh
```diff
@@ -138,5 +136,3 @@ printf '\n]\n' >&3
 exec 3>&-
 mv "$tmp_db" compile_commands.json
 trap - EXIT
-
-echo "[compile_commands.json generated with $entry_count entries]"
```
ericv555: Why was this removed

ericv555 (approved):
Clean.

PR #224, opened 2026-09-16
Refactor HelloRobot::check_safety()

roccoagain:
Split `check_safety()` into small steps, and added safety::Reason flags in safety.hpp so the log shows why safety was triggered. IMO the code is easier to follow now. Shouldn't change robot behavior, apart from the new logging lines.


src/hello_robot.cpp
```diff
@@ -198,61 +198,85 @@ void HelloRobot::update_controls() {
...
+        consecutive_slow_loops = 0;
+        return false;
+    }
+
+    consecutive_slow_loops++;
+    SystemLog.error(Subsystem::GENERAL, "Slow loop with dt: %f, consecutive slow loops: %d\n", dt, consecutive_slow_loops);
+
+    if (consecutive_slow_loops > MAX_CONSECUTIVE_SLOW_LOOPS) {
+        can.issue_safety_mode();
+        SystemLog.error("Kowabunga bitches\n");
```
ericv555, line 244: We are so back

src/hello_robot.cpp
```diff
@@ -198,61 +198,85 @@ void HelloRobot::update_controls() {
...
+        reset_teensy();
+    }
+    return true;
+}
+
+uint8_t HelloRobot::evaluate_safety_reasons(bool is_slow_loop) {
```
ericv555: I think this would be better placed in the safety namespace in safety.hpp. You'll probably want to pass in gimbal power.

src/hello_robot.cpp
```diff
@@ -198,61 +198,85 @@ void HelloRobot::update_controls() {
...
+        reasons |= safety::Reason::GIMBAL_POWER_SETTLING;
+    }
+    return reasons;
+}
+
+void HelloRobot::hold_feeder_position() {
+    float current_feed = (*estimated_state_map)[Cfg::StateName::Feeder].get_position();
+    governor->set_position_reference(Cfg::StateName::Feeder, current_feed);
+    if (has_lower_feeder) {
+        governor->set_position_reference(Cfg::StateName::LowerFeeder, (*estimated_state_map)[Cfg::StateName::LowerFeeder].get_position());
```
ericv555: In Governor, it says not to use the set_position_reference function. Could you create a hold_position or freeze_position function in reference_governor that zeros out a state but sets position. Then in this location you can have something like hold_position(Cfg::StateName::Feeder,current_feed) and also for lower feeder. I feel like that is more idiomatic anyway since, governor should be controlling reference states.

ericv555 (dismissed):
It seems like safety.hpp is being misused here. It was previously for safety_procedures which are fatal errors, but now it is also being used for our safety mode related funtions. I think a more clear refactor would create a SafetyManager class which houses all of your additions to safety.hpp as well as the evaluate_safety_reasons function in hello_robot. I think the additions of reasons is a great change and the use of bitmasks for it is almost certainly the right way to do it. It's just in the wrong file is all.

ericv555 (approved):
We ball and such

PR #225, opened 2026-09-19
Clean up HelloRobot controls function

ericv555:
Moved all send to comms calls out of the update_controls function and put them in a new update_comms() function. It just makes sense...


roccoagain (approved):
Yeah valid

PR #226, opened 2026-09-20
Drain the CAN RX FIFO directly instead of via read()

roccoagain:
`FlexCAN_T4::read()` randomly checks either the FIFO or the mailboxes (basically a coin flip), and since we have no RX mailboxes configured, half its calls return 0 and cut our drain loops short. Calling `readFIFO()` directly drains until the FIFO is actually empty, and in testing gained about 6% more frames per second on Walter.


ericv555 (approved):
We could potentially ball. However, since all motor names are dense integers (i.e they are sequential in an enum) you could potentially replace the motor name map with a motor name array of optional motors/motor pointers. Then you can index that array using the motor name enum. Maybe just food for though though its approved if you wanna merge.

roccoagain:
Will make a note, thanks

